PR #29, opened 2020-04-14
Add a example using requests to readme

mccalluc:


yuanzhou:
@mccalluc thanks for putting effort into the documentation

Issue #53, opened 2020-05-28
Portal application specific index


shirey:
Create a new elasticsearch index set (both public and consortium level) named **portal**.

Create the index based on a transformation provided by @mccalluc.  Pass the transformation function a Python dict as structured to create a json document in the existing **entities** index.  Convert the returned dict to json for storage in the new index set. If an empty dict or None is returned ignore the entry for indexing.  If an exception is raised in the transformation function log the exception description and stack trace and ignore the entry for indexing.

@mccalluc Add your transformation function, named *__transform(document)__*, in the search-api repo in the file /src/elasticsearch/addl_index_transformations/portal.py.  Also, if you want to add logging you should be able to do so by importing the logging library and making calls to logging.info, logging.warning, logging.error, etc...

@yuanzhou  FYI- this will require new Python libraries for inclusion in Docker and to run locally (requirements.txt??) including a git sub-module from the search-schema

yuanzhou:
@shirey I'll take care of the sub-module thing python dependency with docker stuff.

yuanzhou:
@shirey question, once we have the `/<index>/search` in use, do we still support the current `/search` which queries against all indices?

shirey:
@yuanzhou Let's just default /search to the **entities** index.

mccalluc:
I believe this is completed, and it's working well. Thanks! I'll close -- Hope that's ok?

PR #55, opened 2020-05-29
Stub out transform for portal index

mccalluc:
Towards  #53. I'll make a branch from this branch and continue there.


mccalluc:
Close in favor of #56

PR #56, opened 2020-05-29
Mccalluc/add validation and cleaning

mccalluc:
More to come, but this will strip unrecognized properties, and throw an exception if it still doesn't validate. I think this is all I have to contribute for #53, but the issue shouldn't be closed until the index exists.

@yuanzhou :
- It would be nice to be able to grep out all the errors that came from a given run. Would you like to pass in a `batch_id` kwarg?
- <strike>Currently there is no logging, just exception throwing. Let me know if you have any preferences about log formats, and how I will be able to access them.</strike>

Update: Where I've anticipated an error, there's now just logging instead of exception throwing.


yuanzhou:
I think @shirey and @etlds are the right person for reviewing this PR.

shirey:
@mccalluc Per the issue that I created if an exception is thrown it will be logged and the document with the exception will be ignored.  If you want to log anything you can use the standard python *logging* library.

@yuanzhou Before we merge into master please install this on TEST.  You'll need to make sure the submodule defined in .gitmodules is added as well as the libraries in  src/elasticsearch/addl_index_transformations/requirements.txt (eventually, @mccalluc, this requirements.txt file should be merged into the top level src/requirements.txt)

mccalluc:
@yuanzhou : I'm done here for now... Let me know if there's anything else you need from me?

yuanzhou:
I did a localhost build and looks good.

PR #58, opened 2020-06-02
Remove Pipfile

mccalluc:
Is the Pipfile being used for anything? If I'm not going to update it, then it seems like it could go away?


shirey:
I thought this Pipfile is being used for local deployments during development when Docker isn't being used.  I'll let @etlds answer, though..

etlds:
Pipfile is not needed. It is for the python virtual environment. I might check it in by mistake.

mccalluc:
@etlds : If this is good, can you approve, and then I'll merge, unless you prefer a different process?

PR #82, opened 2020-07-07
Update submodule: Pull in typo fixes

mccalluc:
@etlds : 
- Is your preference that I make PRs to devel? 
- Will you do a reindex against all environments, or is that something I should ask for explicitly? (My preference is to be able to rely on the `portal` index being up to date after a PR is accepted... but if that's not feasible, let's talk.)


etlds:
@mccalluc Yes, I think devel is the right branch. Are you talking about setting up an automatic mechanism that kicks off the script one a PR is accepted?
@shirey What do you think?

shirey:
@etlds No, I don't think we want to automate the reindex at this point.  When you accept the PR you should just reindex manually- this is only a single call, right?

PR #85, opened 2020-07-08
Mccalluc/make sure sub objects get mapped

mccalluc:
@tsliaw : Do the `descendant_counts` and `ancestor_counts` look like they'll be sufficient for your needs in the UI? Would it be good to add any other summary stats while we're here? (see https://github.com/hubmapconsortium/portal-ui/issues/672) (ES can [sort of do this natively](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-script-query.html)... but it may be slow, and require features that aren't turned on by default.)
@john-conroy : Does the mapping inside `ancestors` seem good? To confirm, the fields you want are available on the ancestors, they just haven't been mapped? (see https://github.com/hubmapconsortium/portal-ui/issues/662)

@etlds : I'll add you as a reviewer when I've gotten confirmation from my side that this does what we need...

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/__init__.py
+++ b/src/elasticsearch/addl_index_transformations/portal/__init__.py
@@ -16,6 +16,9 @@ from elasticsearch.addl_index_transformations.portal.translate import (
 from elasticsearch.addl_index_transformations.portal.add_everything import (
     add_everything
 )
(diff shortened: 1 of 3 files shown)


mccalluc:
@etlds : Ok, this looks good on our side. Let me know if you see any problems.

PR #105, opened 2020-07-20
Mccalluc/several tweaks

mccalluc:
- @ngehlenborg : I'm having second thoughts about replacing `Unrecognized code` with `[mystery_code]`... I think it would be good to be more clear about about this means... perhaps `Unregistered code: "mystery_code"`? But don't care really strongly.
- @etlds : Do you know if there are other fields where documents might be included recursively? I want to make sure they are all excluded during full-text indexing.
- Any other capitalization fixes besides "daniel Cotter"?


mccalluc:
Yep: That's what's implemented. @etlds : Ok for merge, if it looks ok to you.

PR #116, opened 2020-07-29
special handling for image pyramids

mccalluc:
- Special handling for image pyramids: The mapped field will be a string instead of an array, but not expecting to see any multivalued beyond image pyramids.
- Update submodule. Deployments need to be sure to also update submodule. (But there will be a few more updates... waiting on https://github.com/hubmapconsortium/search-schema/pull/22 and then I'll update the submodule again.)

For my own purposes, I'm recording a version number in the transformed records... would it make any sense to make these numbers within the docker version? If that wouldn't be useful to you, never mind: I've never looked at the docker version number myself, so it's not something that really matters to me.


mccalluc:
(I realized that I made this against master rather than devel. Happy to make a new PR... let me know. Could we make devel the default branch? Would make mistakes like this less likely in the future.)

mccalluc:
Ready to go!

PR #118, opened 2020-07-29
Mccalluc/move version up

mccalluc:
Make the version we put in the index equal the tag (plus a minor-minor version). Do you feel like this is something you'd be comfortable incrementing in your commits?


mccalluc:
Sorry, closing this: Target should be `devel` ... please make that the default branch to prevent errors like this.

PR #119, opened 2020-07-29
Mccalluc/move version up

mccalluc:
Make the version we put in the index equal the tag (plus a minor-minor version). Do you feel like this is something you'd be comfortable incrementing in your commits?


shirey:
@mccalluc If we add a BETA or RC or TEST or some other tag while in development to the end of the version, will this be okay?  You aren't parsing major/minor/minor numbers out of the string, are you?

mccalluc:
@shirey : No, there's no parsing, and I don't care about the internal structure, as long as I know how to increment it: I'd like to continue incrementing on every PR. (For the portal, our docker build script references the VERSION file so nothing gets out of sync: Maybe something like that could work here.)

PR #136, opened 2020-08-10
Mccalluc/no more submodule

mccalluc:
Remove the submodule, and the checked-in derived schemas: Updating one of the enumerations should now only require updating a single yaml file, and that's it. Just updating the yaml could still cause a test failure if there's a syntax problem, or if you happened to touch one of the codes used in the tests, but that should be rare.

There is still code to generate json schema: I would like to keep this in here, and in a future commit I will bring that back as part of validation, but it will be generated on demand and not checked in.

[`search-schema`](https://github.com/hubmapconsortium-graveyard/search-schema) repo is now in the graveyard and archived.


yuanzhou:
Nice! It'll reduce the deployment overhead on my end.

@mccalluc @shirey I think this PR should go into the `devel` branch instead of master. We better test this out on DEV first and maybe also merge into the `1.6` branch and deploy on STAGE?

mccalluc:
> I think this PR should go into the devel

Agreed. Sorry; I'll make a new PR. (Would appreciate having the the default branch set to devel.)

yuanzhou:
> > I think this PR should go into the devel
> 
> Agreed. Sorry; I'll make a new PR. (Would appreciate having the the default branch set to devel.)

I actually asked about making the `devel` as the default branch a while ago. Decision needs to be made by @shirey. Using different branches across different deployment environments does make this a bit hard to keep tracking, and some developers may be confused.

PR #138, opened 2020-08-11
Mccalluc/no more submodule

mccalluc:
Redo #136 against `devel`.


yuanzhou:
@shirey please let me know once you approve and merge this PR into devel branch. Then I can redeploy on DEV. And if you think it can be part of the release, I'll redeploy on STAGE as well.

PR #144, opened 2020-08-18
Mccalluc/gender replaced by sex

mccalluc:

PR #145, opened 2020-08-18
version bump

mccalluc:
Follow-up to #144


yuanzhou:
FYI, the code deployed on STAGE already had bumped the version before reindexing. @shirey  If I merge this PR, then devel would need to be redeployed on DEV and reindexed?

yuanzhou:
The `devel` branch is already on 1.5.2.3 mapper version after the PROD. No need to bump.

PR #146, opened 2020-08-18
Mccalluc/check for version bump

mccalluc:
I want to be required to make a version bump with every PR... (so I don't forget, and have to make more PRs like #145)... but you might not want this in general? I could check that the branch name begins with `mccalluc/`? Or you may have other thoughts...


shirey:
@mccalluc We're using release branches too, so the logic gets more complicated.  Maybe release branches start with "release/", which can be ignored?

yuanzhou:
@mccalluc @shirey the issue that this PR resolves is targeted to the DEV development since it checks against the `devel` branch. I don't think this is very practical because we deploy frequently across different environments. And many times the updates to search-api doesn't require reindexing. When I deploy or redeploy searcha-api, I check with different parties and ask for if reindexing is required, based on which I would bump the mapper version. 

There are also cases that developers would commit directly without using PR, which makes maintenance and keeping tracking the changes hard.

shirey:
@mccalluc I agree with @yuanzhou, though, to me it looks like it skips the devel branch it is still going to wreak havoc if the Pitt team is working across multiple feature, hot fix, bug fix and release branches.

mccalluc:
Sounds like this is a non-starter. Sorry for the bother.

PR #173, opened 2020-09-09
added the sequence_library information

cborromeo:
I needed to add a new specimen type to address: https://github.com/hubmapconsortium/ingest-ui/issues/259.  The new specimen type is 'sequence_library'.  I added the specimen type to the assay_types.yaml file.  @yuanzhou, @shirey asked me to have you merge this change into the devel branch after Nils approves it.  @yuanzhou feel free to delete the cborromeo/sequence-library once you do the merge.


shirey:
@yuanzhou Don't delete the branch until it gets released/merged to master...  @cborromeo FYI, we only delete this type of branch after, at least, it has been merged to a release branch as it could get stuck in devel with stuff that isn't released at the same time.

ngehlenborg (commented):
I am missing context. Is "sequence library" considered a data type? I only know "sequence libraries" as DNA fragments that have been modified to enable next-generation sequencers to "sequence" the DNA.

Is this needed to represent a sample type (like organ piece, etc.)? That's where "sequence library" (better: "sequenc**ing** library") would fit in my mental model. If so, we don't need to add it here, right?

shirey:
@ngehlenborg You're correct, my fault- I pointed Chuck to the wrong file.  This is a "Sample" level type.

@cborromeo Please revert this change in assay_types.yaml and make it in https://github.com/hubmapconsortium-graveyard/search-schema/blob/master/data/definitions/enums/tissue_sample_types.yaml

ngehlenborg:
Thanks @shirey! Was the search-schema repo prematurely archived and moved to https://github.com/hubmapconsortium-graveyard?

shirey:
@ngehlenborg , no again my fault...updated my link.  @cborromeo, I'm going to just do this right now...

shirey (approved):

mccalluc:
> Thanks shirey! Was the search-schema repo prematurely archived and moved to https://github.com/hubmapconsortium-graveyard? @mccalluc did this in August: #136

@ngehlenborg : No, it used to be a separate repo which was referenced as a submodule, and making any tweak to the assay types was a multi-step process. The code that had been there was copied into this repo, the update process is much smoother, and `search-schema` was archived because the separate repo is no longer referenced.

PR #176, opened 2020-09-11
Help with debugging

mccalluc:
Previously `__init__.py` had been executable, but it was kind of buried away. This adds a script at the level that you're usually working, and adds a bit of logging. So, the problem from this morning looks like:
```
$ PYTHONPATH=. elasticsearch/debug.py /tmp/bad-doc.yaml
INFO:root:Begin: Batch unspecified; UUID fa3f5482e3502d7d5d86c61157ee6a7a
DEBUG:root:Mapping single "donor": "None"...
DEBUG:root:... done mapping "donor"
DEBUG:root:Mapping single "origin_sample": {}...
DEBUG:root:... done mapping "origin_sample"
DEBUG:root:Mapping single "source_sample": {}...
DEBUG:root:... done mapping "source_sample"
DEBUG:root:Mapping multi "descendants": {"display_doi": "HBM965.TSLJ.667", "entity_type": ...
DEBUG:root:... done mapping "descendants"
DEBUG:root:Mapping multi "immediate_descendants": {"display_doi": "HBM965.TSLJ.667", "entity_type": ...
DEBUG:root:... done mapping "immediate_descendants"
INFO:root:End: Batch unspecified; UUID fa3f5482e3502d7d5d86c61157ee6a7a
Wrote /tmp/bad-doc.yaml.transformed.yaml
```
- For me, it doesn't work without setting PYTHONPATH... Maybe something is weird in my environment?
- I'm not able to replicate the stack trace you saw; If it's worth pursuing, try running this `debug.py` on it and see how your results compare to mine.


mccalluc:
```
FileNotFoundError: [Errno 2] No such file or directory: 'BUILD'
```
This branch is fine, but when it tries to merge it to master, where `VERSION` is replaced with `BUILD`, you get this this. I could do more to try to fix it... but it may conflict with stuff you already have on `master`. How should we resolve this?

yuanzhou:
That BUILD file needs to be generated. I have put up a script: `https://github.com/hubmapconsortium/search-api/blob/devel/generate-build-version.sh`. You'll need to run that script with Travis prior to starting your test.sh.

That script uses the output from git commands to get the git branch name and the short commit hash ID. Not sure if Travis pulls the code from git.

FYI that BUILD file is ignored by git that's why you don't see it and it needs to be generated after each git clone or git pull so to reflect the current state of the build.

yuanzhou:
@mccalluc @shirey turned out that the failed Travis CI checks are due to the extremely strict code linting tool `Flake8` being used on this `__init__.py`. At one point, even an extra trailing whitespace in a comment line would cause Travis to fail. I fixed the check with this after many tries: https://github.com/hubmapconsortium/search-api/commit/a0197b5ce54a978e996f7fdbbefbb8da39d74754

@mccalluc  we probably should configure Flake8 and make it a bit more tolerant.

PR #178, opened 2020-09-16
Tidy up BUILD version pull

mccalluc:
- It's good not to run too much code just when loading the module, so I've moved it to a function. 
- Changed the print statement to a log to reduce noise on stdout.
- Made the placeholder value more explicit.


yuanzhou:
Looks good! I'll merge and deploy to DEV and STAGE with this new VERSION/BUILD stuff. Then reindex.

PR #179, opened 2020-09-16
Add "set -o errexit"

mccalluc:
Bash's default is not to exit on errors: This makes sense when it's used interactively, but it can make scripts hard to debug. Adding `set -o errexit` will make it exit on any non-zero status, which is probably what you want in a script. (Though there are [differences of opinion](http://mywiki.wooledge.org/BashFAQ/105).)

(Also good to be aware of other options, like `pipefail`.)

### DIFF
--- a/docker/search-api/entrypoint.sh
+++ b/docker/search-api/entrypoint.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -o errexit
 
 # Pass the HOST_UID and HOST_UID from environment variables specified in the child image docker-compose
 HOST_GID=${HOST_GID}
--- a/docker/search-api-docker.sh
+++ b/docker/search-api-docker.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -o errexit
 
 # Print a new line and the banner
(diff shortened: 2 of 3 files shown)


docker/search-api/entrypoint.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -o errexit
 
 # Pass the HOST_UID and HOST_UID from environment variables specified in the child image docker-compose
 HOST_GID=${HOST_GID}
```
mccalluc, line 5: Could be mistaken, but I think this is a no-op: It's already an envvar, so you're just assigning it to itself.

yuanzhou: Not what you thought. The HOST_GID and HOST_UID get passed from system env var to docker-compose, then passed to the docker build. And bing then used by the entrypoint script at container startup. It won’t work if not following this workflow.

yuanzhou:
This entrypoint script is being used by docker build. Any errors occurred would cause the container fail to start. No need to make this PR.

PR #183, opened 2020-09-22
Reorder, add linebreaks between sections

mccalluc:
The list of assays was hard to skim because items had been added on an ad-hoc basis. This just sorts the list by the `description` strings, and add linebreaks between blocks.

@cebriggs7135 : Are you ok with `scATAC-seq (SNARE-seq) [Lab Processed]`, `scRNA-seq (SNARE-seq) [Lab Processed]`, and `seqFISH [Lab Processed]` being used in metadata TSVs? These would be the only terms in metadata TSVs which include processing info in square brackets.

For reference:
```python
from pathlib import Path
from yaml import safe_load, dump

p = Path('src/search-schema/data/definitions/enums/assay_types.yaml')
items = safe_load(p.read_text()).items()
sorted_items = sorted(items, key=lambda item: item[1]['description'])
for i in sorted_items:
  print(f'{i[0]}:')
  print(f'    description: {i[1]["description"]}')
```

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -1,13 +1,25 @@
 AF:
     description: Autofluorescence Microscopy
+
 ATACseq-bulk:
     description: Bulk ATAC-seq
+bulk_atacseq:
+    description: Bulk ATAC-seq [BWA + MACS2]
+
+bulk-RNA:
+    description: Bulk RNA-seq
+salmon_rnaseq_bulk:
+    description: Bulk RNA-seq [Salmon]
+
 CODEX:
     description: CODEX
+codex_cytokit:
+    description: CODEX [Cytokit + SPRM]
+
+image_pyramid:
+    description: Image Pyramid
 IMC:
     description: Imaging Mass Cytometry
-LC-MS-untargeted:
-    description: Untargeted LC-MS
 Lightsheet:
     description: Lightsheet Microscopy
 MALDI-IMS-neg:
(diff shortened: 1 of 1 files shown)


mccalluc:
(Will have a conflict with #185)

mccalluc:
Replaced by #186 -- PR has wrong target.

cebriggs7135:
@mccalluc 

The answer is  YES , I'm "ok with scATAC-seq (SNARE-seq) [Lab Processed], scRNA-seq (SNARE-seq) [Lab Processed], and seqFISH [Lab Processed] being used in metadata TSVs."

cebriggs7135:
@mccalluc Awaiting clarification from Danielle Gutierrez about the best names for mass spec assay_types: 

RE: changing the assay_type enumerations for LC-MS from
 
**| enum | `LC-MS (metabolomics)`, `LC-MS/MS (label-free proteomics)`, or `MS (shotgun lipidomics)` |
 
To
 
| enum | `TMT LC-MS`, `Targeted Shotgun / Flow-injection LC-MS`, or `Untargeted LC-MS` |**

Danielle says: "_The first set of labels could be ok, if that is what the groups are doing, but the bottom set is not clear. Also, the first set of labels doesn’t incorporate all of the groups._ "

mccalluc:
@cebriggs7135 : Thanks for the follow up.
- Note that this PR is closed -- Please add comments to #186  instead.
- Danielle seems to have missed the point that this is a change from one set of labels to the another ("first set of labels could be ok")

Will you be able to get this clarified? Does Phil need to be involved in this converation?

(Do not reply here!)

cebriggs7135:
@mccalluc We can pull Phil in after Danielle/Jeff, who are the LC-MS experts, offer their suggestions.

NOTE: Although Danielle said ""first set of labels could be ok" , she added that "Also, the first set of labels doesn’t incorporate all of the groups." I think she understands we are asking about changing from one set of labels to the other. There are multiple possible combinations of these assays and targets. That's why her solution is to allow all combinations.

PR #186, opened 2020-09-23
Reorder, add linebreaks between sections

mccalluc:
(Replace #183: Had the wrong target.)

The list of assays was hard to skim because items had been added on an ad-hoc basis. This just sorts the list by the `description` strings, and add linebreaks between blocks.

@cebriggs7135 : Are you ok with `scATAC-seq (SNARE-seq) [Lab Processed]`, `scRNA-seq (SNARE-seq) [Lab Processed]`, and `seqFISH [Lab Processed]` being used in metadata TSVs? These would be the only terms in metadata TSVs which include processing info in square brackets.

For reference:
```python
from pathlib import Path
from yaml import safe_load, dump

p = Path('src/search-schema/data/definitions/enums/assay_types.yaml')
items = safe_load(p.read_text()).items()
sorted_items = sorted(items, key=lambda item: item[1]['description'])
for i in sorted_items:
  print(f'{i[0]}:')
  print(f'    description: {i[1]["description"]}')
```

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -1,13 +1,25 @@
 AF:
     description: Autofluorescence Microscopy
+
 ATACseq-bulk:
     description: Bulk ATAC-seq
+bulk_atacseq:
+    description: Bulk ATAC-seq [BWA + MACS2]
+
+bulk-RNA:
+    description: Bulk RNA-seq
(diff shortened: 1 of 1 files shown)


ngehlenborg (approved):

mccalluc:
Remove some reviewers, since this is just a reformatting; Merge changes from master (sc -> sn).

yuanzhou (approved):

PR #187, opened 2020-09-28
Mccalluc/remove underscore metadata

mccalluc:
With [`collectiontype`](https://github.com/hubmapconsortium/portal-ui/issues/1249) now showing up in the metadata, I think we should remind PSC of the understanding we had, that if they need to add fields, they will be prefixed with underscores. I had mentioned this clean-up approach a while ago: You weren't enthused, but I think we need to do something.


ngehlenborg (approved):

mccalluc:
@yuanzhou : Can be merged whenever. Thanks!

yuanzhou (approved):

PR #193, opened 2020-10-02
Add new terms requested by Chris

mccalluc:
Chris believes new terms are necessary to accurately describe these assays, and the lack of the correct terms is [blocking alignment of type terminology](https://github.com/hubmapconsortium/ingest-validation-tools/pull/359#pullrequestreview-498646499).

Chuck has a slight preference just to use the existing terms rather than introducing new ones; Or another approach would be to change some of the descriptions already here, rather than adding new ones.

(Bill prefers that branches be made from master, but this file has diverged significantly, and the merge conflicts would be a mess, so the branch is from devel.)

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/__init__.py
+++ b/src/elasticsearch/addl_index_transformations/portal/__init__.py
@@ -71,9 +71,7 @@ def transform(doc, batch_id='unspecified'):
     ...        }
     ...    }
     ... })
-    >>> del transformed['mapper_metadata']['datetime']
-    >>> del transformed['mapper_metadata']['version']
-    >>> del transformed['mapper_metadata']['validation_errors']
+    >>> del transformed['mapper_metadata']
     >>> pprint(transformed)
     {'ancestor_counts': {'entity_type': {}},
      'ancestor_ids': ['1234', '5678'],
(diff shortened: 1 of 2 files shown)


yuanzhou:
@mccalluc so far we've merged a few PRs into `devel` branch. What's your plan on redeploying onto DEV and STAGE for testing? I've mentioned to @etlds to take a look at this PR since he knows more about the Elasticsearch aspect.

mccalluc:
@yuanzhou: I should have marked this as a draft. I need to get clarity from ngehlenborg and pdblood if this is the way they want to move forward, or if they want to see it find a different approach.

> What's your plan on redeploying onto DEV and STAGE for testing?

My sense was that I don't need to be involved. If there's a new index update on STAGE, happy to look, but there is nothing that should break the portal.

yuanzhou:
@mccalluc we’ll handle the redeploy and reindex. Just let me know when you think these changes need to be deployed on DEV/STAGE to work with the corresponding changes on portal-UI. Otherwise, I’ll wait until we want to release on PROD so we can get things done all together.

mccalluc:
@yuanzhou: Please revert the merge: I need to get confirmation from Nils and Phil that this is the right direction. I apologize for not making this more clear.

ngehlenborg:
So the new terms for `{labeled|label-free}` `{LC-MS|LC-MS/MS}` would ultimately replace the current terms? Do we have a mapping from current terms to new terms and a guarantee that no critical information that is needed for interpretation and processing will get lost in the process? 

I think @cebriggs7135 needs to chime in and I assume that @jswelling might have comments on this as well, given that the ingest infrastructure was written against the original terms.

cebriggs7135:
@nils @jswelling Today, I'll put together a list of the lams already in HubMAP and confirm with each team that one fo the 4 proposed terms is appropriate for those datasets.

PR #195, opened 2020-10-02
Add assay types proposed by Chris

mccalluc:
Replace #193, which was prematurely merged.

Please do not merge until explicit approval is given.

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -20,6 +20,16 @@ image_pyramid:
     description: Image Pyramid
 IMC:
     description: Imaging Mass Cytometry
+
+label-free_lc-ms:
+    description: Label-free LC-MS
+labeled_lc-ms:
+    description: Labeled LC-MS
+label-free_lc-ms-ms:
+    description: Label-free LC-MS/MS
+labeled_lc-ms:
+    description: Labeled LC-MS/MS
+
 Lightsheet:
     description: Lightsheet Microscopy
 MALDI-IMS-neg:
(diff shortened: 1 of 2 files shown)


jswelling (approved):
This is fine.  I might have changed the token order from e.g. label-free_lc-ms-ms to lc-ms-ms_label-free so that they would alphabetize in a more intuitive order.

mccalluc:
(Good suggestion -- There was actually a missing `ms`, so I'm glad to have caught that.)

ngehlenborg (approved):

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -20,6 +20,16 @@ image_pyramid:
...
+
+lc-ms_label-free:
+    description: Label-free LC-MS
+lc-ms_labeled:
+    description: Labeled LC-MS
+lc-ms-ms_label-free:
+    description: Label-free LC-MS/MS
+lc-ms-ms_labeled:
+    description: Labeled LC-MS/MS
+
```
pdblood, line 32: These look good and cover all the MS assay types we currently have. These should replace the existing assay type labels for MS when practical.

pdblood (approved):

pdblood (approved):
I approve the proposed changes, but note that other labels and descriptions in this file still need to be updated -- at least  SNAREseq (SNARE-seq) should be broken up into SNAREseq RNA and SNAREseq AC. It would also be helpful to make the naming of the various seq assays consistent with each other.

mccalluc:
Filed https://github.com/hubmapconsortium/search-api/issues/200 to follow-up.

mccalluc:
@yuanzhou : ready to merge at your convenience.

yuanzhou:
@mccalluc once merged into devel, I'll redeploy on DEV. Then also merge this change into the release branch to be redeployed on STAGE. If all good, we can release to PROD. Does this sound like a good plan?

mccalluc:
@yuanzhou: yes.

PR #196, opened 2020-10-02
Always regen schema

mccalluc:
I believe that running the schema generation as part of the docker build, or running it once and then caching the result is slightly preferable to this, but it's really an implementation detail which is in your court.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -8,9 +8,3 @@ services:
       - HOST_UID=${HOST_UID:-1000}
     init: true
     restart: always
-    volumes:
-      # Mount the VERSION file and BUILD file
-      - "../VERSION:/usr/src/app/src/VERSION"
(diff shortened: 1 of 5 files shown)


yuanzhou:
I'll append some docker tweaks to this PR.

yuanzhou:
Deployed the `mccalluc/always-regen-schema` branch on DEV for testing before merging into `devel`. Let's wait to see if the reindex will go well.

mccalluc:
@yuanzhou : I think generating schemas at build time will work: 
```
$ ./search-api-docker.sh localhost build
...
Successfully tagged search-api:1.8.0

$ docker run --entrypoint ls search-api:1.8.0 src/search-schema/data/generated
README.md
combined-definitions.yaml
dataset.schema.yaml
donor.schema.yaml
sample.schema.yaml
```

Let me know if there's anything else I can do to debug this locally.

PR #204, opened 2020-10-07
Fix lc-ms mistake

mccalluc:
Is this correct now?

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -21,10 +21,8 @@ image_pyramid:
 IMC:
     description: Imaging Mass Cytometry
 
-lc-ms_label-free:
-    description: Label-free LC-MS
-lc-ms_labeled:
-    description: Labeled LC-MS
+ms-ms_label-free:
+    description: Label-free MS/MS
 lc-ms-ms_label-free:
     description: Label-free LC-MS/MS
 lc-ms-ms_labeled:


cebriggs7135 (approved):

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -21,10 +21,8 @@ image_pyramid:
 IMC:
     description: Imaging Mass Cytometry
 
-lc-ms_label-free:
-    description: Label-free LC-MS
-lc-ms_labeled:
-    description: Labeled LC-MS
+ms-ms_label-free:
+    description: Label-free MS/MS
```
pdblood, line 25: The previous version was correct based on my understanding. What is the impetus for this change?

pdblood (changes requested):

src/search-schema/data/definitions/enums/assay_types.yaml
mccalluc: @cebriggs7135?

cebriggs7135: @pdblood 
All mass spec assays are actually 2 step processes that involve an initial MS run on whole/ un-fragmented protein mixture followed by a second MS run on the fragmented  protein mixture
Therefore, every mass spec analysis is formally referred to as “LC-MS/MS” or “MS/MS”, NOT “MS”
 
Liquid chromatography is based on differences in hydrophobicity and since lipids are hydrophobic, LC is not appropriate for lipidomics.

FYI. Metabolites and lipids are too diverse to employ labeling. Hence, metabolomics & lipidomics are always label-free.

The omics type is conveyed in the analyte_class field. Therefore, the words “lipidomics”, “proteomics” & “metabolomics” can be removed from the assay_type.

mccalluc: For what it's worth, "MS/MS" is not found in what I learned last week is the [real list](https://docs.google.com/document/d/1X4wpotexDmvPtUjIN8P5kOgqDGwk0aBLxJSXgSuNqnU/edit).

I don't have a horse in the race, but someone needs to take ownership and make sure we have a clear goal and a consistent process.

ngehlenborg: Is the LC step implicit or does it not exist in this setup?

ngehlenborg (commented):
@pdblood and @cebriggs7135 should agree on this. My input is not needed.

src/search-schema/data/definitions/enums/assay_types.yaml
cebriggs7135: @ngehlenborg  LC is useless for lipids. That's why MS/MS is one of the 3 options. LC separates by hydrophobicity and all lipids are all hydrophobic so there's no way to separate them by LC.

cebriggs7135 (approved):

mccalluc:
Assay names need to by guided by Joel's work. Closing.

PR #205, opened 2020-10-09
Regenerate schemas everytime

mccalluc:


mccalluc:
Sorry, need to fix test error.

yuanzhou:
@mccalluc thanks for making the changes. I'll need to change the permission of that `generate-schemas.sh` so it's executable by the user within the container who runs the python process. But before doing that, I still want to discuss a bit more:

I still don't think regenerating the combined yaml file on each reindex (my understanding is it's getting regenerated either for a full reindex across all docs or reindex for a given UUID/doc) is a good idea, even with using cache.

As @shirey told me earlier, 
> We eventually will start promoting this stuff as open source and want people outside of the Consortium to use it and develop against it, with a complicated setup outside developers will look at it and say “ah, no thanks, these guys make things too complicated, I’ll used something else”.

So  what's your insight if we simply use a single long yaml file and push to github? It avoids all the hassles of this regenerating issue especially we'll add extra tweaks with the docker deployment for permission issue too. By doing this, it still keeps the reindex part separated from the REST search-api (essentially only handles search queries from API consumers against the backend Elasticsearch and returns results). But currently, the reindex part is getting more tight-coupled with the search-api.

mccalluc:
To be clear, I think this what your side requested. If you want something different, that's fine, this can be closed, but the discussion to clarify the right course may not belong on this PR.

Checking built artifacts into version control is in general not a good practice... but if you all want to take over providing the schema, fine with me. I think the "right" place is the Open API spec, but checking it into github is not the end of the world. One challenge is that the schema depends on the enumerations (`assay_types.yaml`, etc.) ... but there also seems to be a push to manage them some different, more ontology-y way. Perhaps you make discrete releases of the schema instead of trying to track every change?

FWIW: 
- I think the privs issues we were having earlier were write-privs: This script does have `a+x`, and I would expect that to be preserved through git, and into docker... If those bits are not preserved, I would like to learn more.
- The repos were separate in the past. I think the primary drive to combine them was to keep it simple to update enumeration. There may be other solutions.

shirey:
@mccalluc I'm not able to see the full change here, where is  _get_schema(doc) called from?  Is the schema generated at run-time or build-time?

mccalluc:
from `_add_validation_errors`... It's taking a document as input, so it's necessarily being called at run-time... because that's what I understood you wanted... but if you want something else, let me know. I've expressed my preferences, but right now I'd just like to get this done.

PR #230, opened 2021-01-04
Get endpoint from context

mccalluc:
For the indexing work, I'll need access to the Entity API url config. What is the right way to do this?
- This PR wraps my transform in `app.app_context()`... but maybe as you'll be calling it, the context would already be available, and what I really need to do is tweak the tests?
- Alternatively, I could read the config directly, rather than going through flask?
- Or maybe I should access it though the `Indexer` object?
- Additional argument to `transform()`?

Making this a draft, just to put something out there, but I'd be surprised if I'm on the right track.


yuanzhou:
- The Entity API URL is configured as https://github.com/hubmapconsortium/search-api/blob/test-release/src/instance/app.cfg.example#L14 and in the code, you can call `current_app.config['ENTITY_WEBSERVICE_URL']`
- Looking at the indexer, the Entity API URL gets passed into the constructor: https://github.com/hubmapconsortium/search-api/blob/test-release/src/elasticsearch/indexer.py#L32

To me, either way is fine.

yuanzhou:
From Chuck: 
> I may come back to this — the idea is to let indexing provide access to information beyond the input doc — but if it’s in your way, feel free to close, but keep the branch, and I might reopen the PR later.

Will just leave the branch and PR here.

mccalluc:
No longer a priority. Closing.

PR #240, opened 2021-03-04
parse partonomy

mccalluc:
No change to indexing right now... just a demo of how I plan to use the partonomy, before this code gets entangled with everything else.

@shirey :
- Do you feel comfortable with including the partonomy this way? ... or would it be better not to write to a gitignored directory? Could just check it in, or possibly use a submodule, but this seems simpler and clearer.

@bherr2 :
- Are you comfortable with how I'm parsing the data, and the structural assumptions I'm making? ie, one parentless root node, assume there is only one label and parent on each node, etc.

Towards: https://github.com/hubmapconsortium/portal-ui/issues/1572

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/demo_partonomy.py
+++ b/src/elasticsearch/addl_index_transformations/portal/demo_partonomy.py
@@ -1,0 +1,10 @@
+import requests
+from pathlib import Path
+
+partonomy_path = Path(__name__).parent / 'cache/partonomy.jsonld'
+if not partonomy_path.exists():
+  partonomy_url = 'https://cdn.jsdelivr.net/gh/hubmapconsortium/hubmap-ontology@1.0.0/ccf-partonomy.jsonld'
+  partonomy_path.write_text(requests.get(partonomy_url).text)
+
+print(partonomy_path.read_text())
+
--- a/src/elasticsearch/addl_index_transformations/portal/demo_partonomy.py
+++ b/src/elasticsearch/addl_index_transformations/portal/demo_partonomy.py
@@ -1,10 +1,32 @@
 import requests
 from pathlib import Path
+from json import loads
 
-partonomy_path = Path(__name__).parent / 'cache/partonomy.jsonld'
-if not partonomy_path.exists():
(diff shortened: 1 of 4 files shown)


bherr2 (approved):
Makes sense and looks good!

mccalluc:
Removing Bill has reviewer: @yuanzhou as codeowner should be sufficient.

src/elasticsearch/addl_index_transformations/portal/demo_partonomy.py
```diff
@@ -0,0 +1,60 @@
...
+
+def build_tree_index():
+    '''
+    Returns a tuple:
+        - A tree, where each node has value, parent_id, and children.
+        - A dict indexing into every node of the tree by id.
+    '''
+    partonomy_path = Path(__file__).parent / 'cache/partonomy.jsonld'
+    if not partonomy_path.exists():
+        partonomy_url = 'https://cdn.jsdelivr.net/gh/hubmapconsortium/hubmap-ontology@1.0.0/ccf-partonomy.jsonld'
```
bherr2, line 14: Switch URL to https://cdn.jsdelivr.net/gh/hubmapconsortium/hubmap-ontology@1/ccf-partonomy.jsonld so that it gets the latest version (a new version may appear at end of March but definitely end of May).

bherr2 (changes requested):

src/elasticsearch/addl_index_transformations/portal/demo_partonomy.py
mccalluc: @bherr2 : I don't think that's the behavior we want: I really prefer to pin dependencies and have the behavior of code be consistent until we make an explicit change.

If this doesn't seem correct to you, let's talk on the call this afternoon.

bherr2: Your call. Not how I'd do it, but I understand your reasoning.

bherr2 (approved):

mccalluc:
Branch extended and new PR filed: #242. Closing.

PR #263, opened 2021-03-26
Make the message more explicit when the translation is missing

mccalluc:
I had totally forgotten where the curly braces come from. My sense is that:
- We identified a small gap in the process, and it will become clear who has responsibility for supplying missing translations going forward.
- If "DART-FISH" is "DART-FISH", great: It just needs to be explicit.
- It's not the end of the world for these to show up sometimes on unpublished data ... Erroring out would be too aggressive?
- Nothing else I need to do here.

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/translate.py
+++ b/src/elasticsearch/addl_index_transformations/portal/translate.py
@@ -11,7 +11,7 @@ class TranslationException(Exception):
 
 
 def _unexpected(s):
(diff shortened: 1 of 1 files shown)


yuanzhou (approved):

khanshawPSC (approved):

yuanzhou:
I’ll merge after releasing couple of things being tested on TEST using the “test-release” branch early next week.

PR #265, opened 2021-03-29
mapping config

mccalluc:
Fix https://github.com/hubmapconsortium/portal-ui/issues/1665

Plan:
- Check in a configuration file as YAML, and make it accessible at `get_config`.
- Tweak config to get the indexing behavior the portal needs.
- Add tests of indexing behavior against a live ES instance with this config.

@yuanzhou : While there are some configurations that can be applied to a pre-existing index, I think things will work best if we assume a clean slate. You will need a blue-green approach to swapping indexes on production.

Filing as draft: I'm going to work on the live ES testing, but I wanted to get this on your radar.

-----------------

update: I think this is ready to go. One question: I remember there was something about [date detection](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html#date-detection) that needed to be tweaked... I don't know where that tweak was made, but it would be good to capture that in the mapping here, and add a test -- Do you remember the details of that?


mccalluc:
@yuanzhou : Tests are failing on Travis... before I get too deep into this wanted to check:
- Does this seem like a good idea, in the abstract?
- What version of Elasticsearch is being used on the server?
- If you install ES locally, do the tests work for you?

Tests pass locally for me, with a brew-installed Elasticsearch.

yuanzhou:
@mccalluc to answer your questions:

> Does this seem like a good idea, in the abstract?

It's unclear to me what issue this mapping will address or what is the major benefit of introducing it...

> What version of Elasticsearch is being used on the server?

We are currently using Amazon Elasticsearch Service. Elasticsearch version 7.4 on DEV/TEST/STAGE/PROD. PROD has more nodes across several availability zones compared to the rest.

> If you install ES locally, do the tests work for you?

I've used ES docker locally long time ago, but we've been using the AWS ES for testing purposes.

For the failed travis test, it looks like you'll need to create the index first before adding a new doc.

For the date detection, we use the following when creating a new index:
````
index_info_dict = {
                "settings": {
                    "index" : {
                        "mapping.total_fields.limit": 5000,
                        "query.default_field": 2048
                    }
                },
                "mappings": {
                    "date_detection": False
                }
            }
````

**Comment by mccalluc** [comment]
Thanks for the pointers -- I'll do some work on this, and probably ping you later today.

> It's unclear to me what issue this mapping will address or what is the major benefit of introducing it...

Two main areas:
- More efficient: For free text search, instead of needed to copy every value to the `everything` field, which is stored as part of the source document, we can use `copy_to` and an index-only field. Also, since we never sort by relevance, we could get rid of [norms](https://www.elastic.co/guide/en/elasticsearch/reference/current/norms.html) and save more disk.
- More features: We might enable search-as-you-type first, others in the future.

> Elasticsearch version 7.4

Thanks!

> but we've been using the AWS ES for testing purposes

For unit testing, I think it would make sense to just use a local ES instance: Don't need to worry about passing credentials to Travis, or accumulating indexes that were only for test purposes. For integration testing, AWS ES certainly has a role!

> For the failed travis test

Thanks: re-reading this, I'm thinking I might need to add another `?refresh` to get synchronous behavior.

> For the date detection, we use the following when creating a new index:

Thanks! I'll add that to the config I've started in this PR.

**Comment by mccalluc** [comment]
@yuanzhou -- With tests passing, this much is good, but before it is merged, I would like to tweak `es_writer.py` so that it pulls in the new configuration in `config.yaml` ... Should I do this, or would you like to?

File: src/elasticsearch/addl_index_transformations/portal/__init__.py [hunk 1]
```diff
@@ -40,6 +40,16 @@ def _get_version():
     return 'no-build-file'
 
 
+def get_config():
```
mccalluc, line 43: Where it matters, we're just reading the yaml rather than using this method. If just referencing this file is fine, can get rid of this method.

src/libs/es_writer.py
```diff
@@ -100,3 +91,22 @@ def create_index(self, index_name):
             msg = "Exception encountered during executing ESWriter.create_index()"
             # Log the full stack trace, prepend a line with our message
             logger.exception(msg)
+
+
+def _get_configuration(index_name):
+    if index_name == 'portal':
```
yuanzhou, line 97: Currently the `indexer.py` is probably the only place that uses the `es_writer.py` and when it creates the indices, there's no such index named `portal`. The actual ones are `hm_public_portal` and `hm_consortium_portal`, but from the end user perspective, they'll see `portal` as the index name because we are not exposing the actual real index names to the end users.

Was your intention to use the mapping from `config.yaml` when  `hm_public_portal` and `hm_consortium_portal` are being created?

yuanzhou: My understanding of the `config.yaml` usage is for creating the portal indices, if this is correct, I'm also thinking that we should move this loading of the mapping details to the `indexer.py`. Because the `indexer.py` reads in the indices configuration from https://github.com/hubmapconsortium/search-api/blob/test-release/src/instance/app.cfg.example#L29 and creates the indices at https://github.com/hubmapconsortium/search-api/blob/test-release/src/elasticsearch/indexer.py#L73. The `es_writer.py` was designed as a library and it shouldn't rely on reading a `config.yaml` for creating certain indices. 

As a result, we can add another parameter to the `ESWriter.create_index(index_name, config)` and the `config` would include the settings and mappings definition, which can be passed in at runtime. Eventually, we can specify the settings and mappings in `indexer.py` for both portal indices and the other original indices. If someone else wants to use the `es_writer.py`, they'll be able to pass in their own configurations for creating indices easily without making changes to the `es_writer.py`. 

@shirey @mccalluc does this sound like a good move?

shirey: @mccalluc @yuanzhou and I just discussed this, what Zhou is saying seems correct- these configuration changes only need to happen when the indices are created.  Our indices are currently created when we kick off a full-reindex with a standalone script, but we could separate that piece if helpful to you during development.

mccalluc: > Was your intention to use the mapping from config.yaml when hm_public_portal and hm_consortium_portal are being created?

Assuming these are the instances that have been run through `src/elasticsearch/addl_index_transformations/portal`, yes.

> As a result, we can add another parameter to the ESWriter.create_index(index_name, config) and the config would include the settings and mappings definition, which can be passed in at runtime. ... does this sound like a good move?

As far as I follow it, yes.

> these configuration changes only need to happen when the indices are created.

yes.

>  Our indices are currently created when we kick off a full-reindex with a standalone script, but we could separate that piece if helpful to you during development.

I'm not sure that that is necessary: I think the multiple iterations needed for the recent nested-metadata PR were exceptional.

@shirey @yuanzhou : Here are the next steps as I see them:
- Chuck will make a new PR, branching from just before the last commit.
- Zhou will review, and if appropriate, merge.
- Zhou or Bill will do what is needed so that the mappings Chuck specifies are applied when a new index is set up.
- Chuck will have occasional mapping updates, and will be able confirm some aspects of the functionality with `test-es.py`... but other issues may not be apparent without pointing a portal instance at the result of a reindex... Using TEST or DEV for this works for me... but it might not work for you, in which case there should be guidance on setting up a local instance.
- At some point, the PROD index is updated. I believe a microservices architecture should mean that extended downtime is not necessary... but I don't think that's the case here.

mccalluc:
Replaced by #282 (branch from before last commit.)

PR #266, opened 2021-03-29
Mccalluc/metadata metadata

mccalluc:
This would fix https://github.com/hubmapconsortium/portal-ui/issues/1466. It feels kludgy... but since dev-search is a developer-only feature, I don't think it makes sense to create the complex `must` and `should` ES queries which select the records we need with the data in its current shape.


yuanzhou:
@mccalluc how do you plan to verify this change? I can merge and redeploy on TEST with a reindex, would you also need to redeploy portal-ui to test?

mccalluc:
@yuanzhou :
> would you also need to redeploy portal-ui to test?

No: What I'm hoping is that after the reindex on test, on dev-search (https://portal.test.hubmapconsortium.org/dev-search?has_metadata=true) it will show some Samples and Donors under the "Entity Type" facet at the top, which it doesn't right now.

PR #270, opened 2021-03-30
Follow-up to #266: Get the case right

mccalluc:


yuanzhou:
Still failed after redeployment and reindex. Lots of errors like below:

````
ERROR in es_writer: Failed to write doc of uuid: a7d0202b32ede8613b1d38f28d9bbb0a to index: hm_consortium_portal
ERROR in es_writer: Error Message: {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"object mapping for [metadata.metadata] tried to parse field [metadata] as object, but found a concrete value"}],"type":"mapper_parsing_exception","reason":"object mapping for [metadata.metadata] tried to parse field [metadata] as object, but found a concrete value"},"status":400}
````

PR #282, opened 2021-04-02
Mccalluc/mapping config revert

mccalluc:
Replaces #265 -- Branch from just before the last commit, which took it in the wrong direction. Next steps, as outlined in that issue:

>-    Zhou will review, and if appropriate, merge.
>-    Zhou or Bill will do what is needed so that the mappings Chuck specifies are applied when a new index is set up.
>-    Chuck will have occasional mapping updates, and will be able confirm some aspects of the functionality with test-es.py... but other issues may not be apparent without pointing a portal instance at the result of a reindex... Using TEST or DEV for this works for me... but it might not work for you, in which case there should be guidance on setting up a local instance.
>-    At some point, the PROD index is updated. I believe a microservices architecture should mean that extended downtime is not necessary... but I don't think that's the case here.


yuanzhou:
Thanks @mccalluc! If I understand correctly, I'll make the corresponding changes to grab the settings and mappings definition from your `config.yaml` to create the portal indices? So down the road, when you make tweaks to the yaml config, your test script and the `indexer.py` should both pick up those updates when recreating the portal indices.

mccalluc:
>  So down the road, when you make tweaks to the yaml config, your test script and the indexer.py should both pick up those updates when recreating the portal indices.

yes, that sounds right. As we go forward, it would help me to have some clear expectations about how and when changes to the mappings can make it to production. I don't think they will be frequent... but if the delivery time to production is uncertain / longer than a week, that could slow down other parts of development.

yuanzhou:
FYI @mccalluc I've made the corresponding changes and appended to your branch. I created another PR #283 to merge into `dev-integrate` first and test with a full reindex on DEV to make sure all looks good before merging the PR #282 into `test-release`.

mccalluc:
@yuanzhou : The default behavior we've been relying on creates a keyword subfield for every field. The commits I've pushed add that to the dynamic template that was added earlier.

yuanzhou:
@mccalluc is this the cause to the error reported at the bottom of this DEV deployment? https://github.com/hubmapconsortium/search-api/pull/283

I'll merge your new fixes to `dev-integrate` and test again.

yuanzhou:
@mccalluc OK, I just merged your new commits into `dev-integrate` and redeploy on DEV, started a full reindex. The portal-ui error from yesterday is gone now. I'll have to wait until the reindex is completed to be 100% sure. It looks good so far.

mccalluc:
@yuanzhou - Great! When this is merged to `test-release` I think the next thing I'll try is getting rid of norms: As I understand it, those are only used when sorting by relevance, but we never do that: We're always sorting by some field. Getting rid of them will save some disk space.

Thanks for tolerating this iterative process: If there's more I should be doing on my end, let me know.

PR #283, opened 2021-04-03
Mccalluc/mapping config revert

yuanzhou:
Testing on DEV before merging the PR #282 into `test-release` and deploy on TEST.


yuanzhou:
@mccalluc  I redeployed on DEV and the full reindex went well. But portal-ui couldn't display the `/dev-search` page:

image

PR #292, opened 2021-04-15
test highlighting and removing norms

mccalluc:
Towards https://github.com/hubmapconsortium/portal-ui/issues/1718 and https://github.com/hubmapconsortium/portal-ui/issues/1719

I went into this thinking that we could remove norms because we never sort by relevance in the portal... but it doesn't seem to make a different in the results ... either:
- (a) I'm not doing it correctly,
- (b) or I'm doing it correctly, and it's an optimization, but doesn't change the results,
- (c) or, it could change the results, but I'm not testing the right thing.

I think it's (b), but could be wrong.

I'd suggest looking at the file in its final state, rather than trying to read the diff.

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/test_es.py
+++ b/src/elasticsearch/addl_index_transformations/portal/test_es.py
@@ -113,6 +113,8 @@ def test_highlight():
         'query': {'match': {'all_text': 'term'}},
         'highlight': {'fields': {'*': {}}}
     })
+    assert sorted(search_response['hits']['hits'][0].keys()) \
+        == ['_id', '_index', '_score', '_source', '_type', 'highlight']
     assert [
         hit['highlight']['all_text']
         for hit in search_response['hits']['hits']
--- a/src/elasticsearch/addl_index_transformations/portal/test_es.py
+++ b/src/elasticsearch/addl_index_transformations/portal/test_es.py
@@ -39,19 +39,17 @@ def query_docs(query):
 
 
 def test_tokenization_and_search():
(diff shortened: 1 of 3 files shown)


mccalluc:
`KeyError: 'highlight'` on travis, but not locally. Changing to draft.

src/elasticsearch/addl_index_transformations/portal/test_es.py
```diff
@@ -1,96 +1,126 @@
...
-    get_index_response = requests.get(f'{base_url}/{index}').json()
-    LOGGER.info(f'get index:\t{get_index_response}')
+    # get_index_response = requests.get(f'{base_url}/{index}').json()
     # assert 'dynamic_templates' in get_index_response[index]['mappings']
 
 
+def index_docs(docs):
+    for i, doc in enumerate(docs):
+        # NOTE: Without '?refresh', the index is not guaranteed to be
+        # up-to-date when the response returns. Should not be unused
```
john-conroy, line 32:
```suggestion
        # up-to-date when the response returns. Should not be used
```
Unused or used?

john-conroy (commented):
Should you wait to merge until you better understand/observe the effect of removing norms?

mccalluc:
> Should you wait to merge until you better understand/observe the effect of removing norms?

The primary benefit would be reduced disk use. We don't use relevance sorting in the portal, so I think it's low risk. What I can do is see if the ES administrative APIs give us information on disk use, and that would be a way for me to check my hypothesis in the context of these tests.

PR #310, opened 2021-04-28
update assays

### LINKED ISSUE #309: add entries to assay_types.yaml for unmapped names
replaces #268. Joel says:
> ...  for now the assay type name service is being served from search-api . ... Updating an assay_type in assay_types.yaml within search-api is intended to be the authoritative change.

mccalluc:
Fix #309.

Three assay types are showing up as untranslated when I log in:
```
No translation for "Slide-seq"
  19 datasets
No translation for "DART-FISH"
  2 datasets
No translation for "salmon_rnaseq_10x_sn"
  1 datasets
```

- For "Slide-seq" and "DART-FISH", googling indicates that that is the common capitalization.
- Those two were also marked as "primary" where they were commented out, while I took `salmon_rnaseq_10x` as a model for `salmon_rnaseq_10x_sn`, so it is not "primary" ... but I don't know if any of these are correct.
- For `salmon_rnaseq_10x_sn` I introduce `is_sn` as a vitessce-hint. The others have no vitessce-hints, because there were none when they were commented out ... but again not sure if this is correct.


src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -207,6 +207,14 @@ salmon_rnaseq_10x:
     primary: false
     vitessce-hints: ['is_sc', 'rna']
 
+salmon_rnaseq_10x_sn:
+    description: snRNA-seq (10x Genomics) [Salmon]
+    alt-names: []
+    # Assuming false, based on salmon_rnaseq_10x?
+    primary: false
+    # First use of "is_sn": Generalize vitessce config generation?
+    vitessce-hints: ['is_sn', 'rna']
```
mccalluc, line 216:
```suggestion
    primary: false
    vitessce-hints: ['is_sn', 'rna']
```

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -287,11 +293,11 @@ salmon_sn_rnaseq_10x:
...
-# Slide-seq:
-#     description: ????
-#     alt-names: []
-#     primary: true
-#     vitesse-hints: [[]
+Slide-seq:
+    description: Slide-seq
+    alt-names: []
+    primary: true
+    vitesse-hints: []
```
ilan-gold, line 300: I think we can add `rna` here since this is spatially resolved rna-seq

mccalluc, line 300:
```suggestion
    vitesse-hints: ['rna']
```

mccalluc: fixed!

ngehlenborg:
Removed myself as reviewer to prevent this from being held up.

yuanzhou (approved):
Looks good to me, and I can get this out on TEST once all approved and merged.

jswelling (changes requested):
DART-FISH and Slide-seq were commented out because we didn't have an official "description" entry for them, but clearly we need those entries now.
salmon_rnaseq_10x_sn is actually a variant spelling of salmon_sn_rnaseq_10x, which is already defined- the version with _sn at the end is more consistent but we already have datasets with the latter spelling.  I think the right solution there is to make salmon_rnaseq_10x_sn an alt-name of salmon_sn_rnaseq_10x.

mccalluc:
@jswelling -- Think I've got it correct, now. Sorry not to be clear on this, but would there be any functional difference between these two?:
```
xyz:
    alt-names: ['abc']
```
```
abc:
    alt-names: ['xyz']
```
Does it really make any difference which is the primary name?

jswelling:
Yes- the difference is that the top name is the 'real' name and the names in the alt-name list are 'canonicalized' to the real name.  So for example given 
> xyz:
>    alt-names: ['abc']
a lookup for either 'xyz' or 'abc' will yield 'xyz'.  No lookup will ever yield 'abc'; the translation is meant to be one-way.

jswelling (approved):

mccalluc:
@yuanzhou -- Zhou -- I'm sorry, you usually do the merges here. Would you like to protect the test-release branch to prevent me from doing this in the future?

PR #311, opened 2021-05-02
The sc_atac_seq_sci datatype is derived from sciATACseq and not 'prim…

jswelling:
…ary'

assay_types.yaml controls the definitions of assay types and datasets derived from assays.  The entry for sc_atac_seq_sci was incorrectly marked as primary.

### DIFF
--- a/examples/libs/assay_type_out.txt
+++ b/examples/libs/assay_type_out.txt
@@ -60,6 +60,7 @@ non-primary names:
 - salmon_rnaseq_sciseq
 - salmon_rnaseq_snareseq
 - salmon_sn_rnaseq_10x
+- sc_atac_seq_sci
 - sc_atac_seq_snare
 - sc_atac_seq_snare_lab
 - sc_rna_seq_snare_lab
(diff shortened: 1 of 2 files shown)


jswelling:
Travis problem fixed.

yuanzhou (approved):

mccalluc (commented):
@jswelling -- Should this change have any implications on the front end? Just searching "primary" in `addl_index_transformations/` and `portal-ui`, I don't think I see any references to this. Are you aware of other code that hits this attribute, and maybe the front-end depends on that dependency? (Nils may not have time to respond to this PR himself.)

jswelling:
@mccalluc changing sc_atac_seq_sci away from 'primary' should cause it to not be displayed on the 'Data Submission' page as an option.  If you look there now you will see both sciATAC-seq and sciATAC-seq [SnapATAC]; the latter is a manifestation of the incorrect reporting of sc_atac_seq_sci as primary.  That's the only visible impact of this change I'm aware of.

mccalluc (approved):

PR #314, opened 2021-05-11
Remove references to "everything"

mccalluc:
Fix https://github.com/hubmapconsortium/portal-ui/issues/1818

Follow-up to https://github.com/hubmapconsortium/portal-ui/pull/1809


src/elasticsearch/addl_index_transformations/portal/__init__.py
```diff
@@ -182,6 +161,10 @@ def _map(doc, clean):
     # The recursion is usually not needed...
     # but better to do it everywhere than to miss one case.
     clean(doc)
+
+    single_valued_fields = ['donor', 'origin_sample', 'source_sample', 'rui_location']
+    multi_valued_fields = ['ancestors', 'descendants', 'immediate_ancestors', 'immediate_descendants']
```
john-conroy, line 166: We were just importing this from elsewhere before. No impactful change?

john-conroy (approved):

src/elasticsearch/addl_index_transformations/portal/__init__.py
mccalluc: yep, no change: The outside file now doesn't do anything else now, so makes sense to inline.

mccalluc:
> I don't think I know enough about the underlying logics. I'll approve and we can find out if it addresses the issue correctly after redeploying and reindex on TEST.

Zhou: My hope is that this will reduce the disk use of the index a bit: If it would be easy for you get a before and after number, that would be great, but don't worry if that's not feasible.

yuanzhou:
@mccalluc it's good to reduce the disk usage. I don't think it'll be easy for me get the before/after number because we are using AWS elasticsearch cluster for PROD, I'll see what I can do. In general the disk space is not a concern, we can always increase if needed.

I'll merge this PR and schedule a redeployment is no more changes.

yuanzhou:
Redeployed and reindexed on TEST. Here is the before/after disk usage comparision:

| index                | Before (Size in bytes) | After (Size in bytes) |
|----------------------|------------------------|-----------------------|
| hm_public_portal     | 130.75 MiB             | 115.32 MiB            |
| hm_consortium_portal | 254.41 MiB             | 228.48 MiB            |

PR #317, opened 2021-05-13
Welling/regularize snare assay types

jswelling:
This PR updates and regularizes SNARE assay names and descriptions.  There is a companion PR for ingest-validation-tools.

PR #320, opened 2021-05-17
Mccalluc/still no translation

mccalluc:
Towards https://github.com/hubmapconsortium/portal-ui/issues/1761. (Waiting to call it fixed until I see the fix in the portal.)

@jswelling -- The fix is one line at the bottom: `for name in v['alt-names'] + [k]`. Is this ok, or should I be going through something else rather than reading the yaml directly?


jswelling:
@mccalluc the intention would be to import the class AssayType from libs, and then do something like the following.  The problem is that there is no easy way to reference the libs directory from translate.py.
```
def _data_types_map(ks):
    assert len(ks) == 1 or (len(ks) == 2 and 'image_pyramid' in ks)
    single_key = ks[0] if len(ks) == 1 else ks
    try:
        return AssayType(single_key).definition
    except RuntimeError:
        if isinstance(single_key, list):
            return _unexpected(' / '.join(single_key))
        else:
            return _unexpected(single_key)
```

The assertion is true for the current usage pattern, since 'vis lifting' of image pyramids is the only known case where a dataset has two datatypes.

mccalluc:
@jswelling 
>  The problem is that there is no easy way to reference the libs directory from translate.py

No, that doesn't seem to be a problem: The import just works. On the other hand, there is a problem with the mapping: I expect
```
['image_pyramid', 'AF']
```
to produce
```
['Autofluorescence Microscopy [Image Pyramid]']
```
but instead it gives
```
['No translation for "image_pyramid / AF"']
```

It seems like there's an analogous case that you've checked in to `assay_type_out.txt`: I think the behavior just is not correct, and the fix needs to be `assay_type.py`
```
['PAS', 'Image Pyramid'] (this is a complex alt name) -> exception No such assay_type ['PAS', 'Image Pyramid'], even as alternate name
```

I could try tackling it, but wanted to check in with you first.

jswelling:
@mccalluc the problem is actually in assay_types.yaml, where the alt-names linking ['Image Pyramid', 'AF'] as an alt-name was commented out.  I have a fix in the branch welling/add_pii_to_assay_type but I can't usefully create a PR until https://github.com/hubmapconsortium/search-api/pull/317 is merged.  I think we've reached stability on those names so hopefully that merge will happen in the next day and you'll be able to see what I'm talking about.

mccalluc:
>  I have a fix in the branch welling/add_pii_to_assay_type but I can't usefully create a PR until #317 is merged.

@jswelling -- That PR is merged... What's the next step on this PR?

jswelling:
@mccalluc the PR that should correct this has been merged into the dev-integrate branch but I don't know how to trigger a Travis test against that branch rather than the one associated with this PR.

> @jswelling
> 
> > The problem is that there is no easy way to reference the libs directory from translate.py
> 
> ...On the other hand, there is a problem with the mapping: I expect
> 
> ```
> ['image_pyramid', 'AF']
> ```
> 
> to produce
> 
> ```
> ['Autofluorescence Microscopy [Image Pyramid]']
> ```
> 
> but instead it gives
> 
> ```
> ['No translation for "image_pyramid / AF"']
> ```
> 
> It seems like there's an analogous case that you've checked in to `assay_type_out.txt`: I think the behavior just is not correct, and the fix needs to be in `assay_type.py`
> 
> ```
> ['PAS', 'Image Pyramid'] (this is a complex alt name) -> exception No such assay_type ['PAS', 'Image Pyramid'], even as alternate name
> ```
> 
> I could try tackling it, but wanted to check in with you first.

mccalluc:
Next step for me: Take the work in this branch and either re-root it or just re-do it on `dev-integrate`.

PR #326, opened 2021-05-24
Welling/add pii to assay type

jswelling:
This PR adds two new boolean properties, vis_only and contains_pii, to the assay_type data structure and assigns values for those fields to the existing assays.  

* vis_only is true if the data type contains only visualization information, and thus should be subject to vis lifting.  It defaults to False.  It exists to support future extensions to vis lifting, and because other changes make the presence of "image pyramid" in the data type list an unreliable indicator.
*  contains_pii is true if the data type contains personally identifying information and thus must be protected.  It defaults to True for safety.  

The AssayType class and the entries in assay_types.yaml are extended to support these new properties.


src/app.py
```diff
@@ -76,7 +74,7 @@ def assaytypes():
     primary = None
     simple = False
     for key, val in request.args.items():
-        print(f'{key}:{val}')
+        #print(f'{key}:{val}')
```
mccalluc, line 77: Delete, or change to an `info` or `debug` log?

src/libs/assay_type.py
```diff
@@ -81,14 +81,22 @@ def __init__(self, name: StrOrListStr):
         self.vitessce_hints = (this_def['vitessce-hints']
                               if 'vitessce-hints' in this_def
                               else [])
+        self.contains_pii = (this_def['contains-pii']
+                             if 'contains-pii' in this_def
+                             else True)  # default to True for safety
+        self.vis_only = (this_def['vis-only']
+                             if 'vis-only' in this_def
+                             else False)  # False is more common
```
mccalluc, line 89:
```suggestion
        self.contains_pii = this_def.get('contains-pii', True)  # default to True for safety
        self.vis_only = this_def.get('vis-only', False)  # False is more common
```
more concise

src/libs/assay_type.py
```diff
@@ -118,14 +126,32 @@ def main() -> None:
...
     """
-    for name, note in [('codex', 'this should fail'),
-                       ('CODEX', 'this should work'),
-                       ('codex_cytokit', 'this is not primary'),
-                       ('salmon_rnaseq_bulk', 'this is an alt name'),
-                       (['PAS', 'Image Pyramid'],
-                        'this is a complex alt name'),
-                       (['IMC', 'foo'],
-                        'this is an invalid complex alt name')]:
+    cases = [('codex',
```
mccalluc, line 129: This testing approach is getting awkward. It essentially splits the assertions between two files: this one, and the output that you diff. Problems noticed in the diff would need to be found here and corrected. Suggest using doctests or pytest.

src/libs/assay_type.py
jswelling: Agreed; much better.

src/app.py
jswelling: Yeah, delete.

src/libs/assay_type.py
jswelling: I've converted it to doctest.  I'd thought that doing so would produce something that was unacceptably verbose, but it's not really that bad, and as you say the unification of the test prototype and the code itself has merit.

src/libs/assay_type.py
```diff
@@ -30,6 +30,49 @@
 class AssayType(object):
     """
     A class intended to represent a single assay type, derived or otherwise.
+
+y    >>> AssayType('codex')  # should be uppercase
```
mccalluc, line 34:
```suggestion
    >>> AssayType('codex')  # should be uppercase
```
With the extra character, I'm guessing this wasn't being seen by doctest.

src/libs/assay_type.py
```diff
@@ -30,6 +30,49 @@
...
+    ['image_pyramid', 'Image Pyramid', True, False]
+
+    >>> a_t = AssayType(['Image Pyramid', 'foo'])  # invalid complex alt name
+    Traceback (most recent call last):
+    ...
+    RuntimeError: No such assay_type ['Image Pyramid', 'foo'], even as alternate name
+
+    >>> a_t = AssayType('salmon_rnaseq_10x_sn')  # simple valid alt name
+    >>> [a_t.name, a_t.description, a_t.vis_only, a_t.contains_pii]
+    ['salmon_sn_rnaseq_10x', 'snRNA-seq [Salmon]', False, False]
```
mccalluc, line 74: Helper function makes it easier to focus on what matters:
```suggestion
    >>> def name_desc_vis_pii(assay):
    ...     a_t = AssayType(assay)
    ...     return [a_t.name, a_t.description, a_t.vis_only, a_t.contains_pii]

    >>> name_desc_vis_pii('CODEX')
    ['CODEX', 'CODEX', False, False]

    >>> name_desc_vis_pii('codex_cytokit')  # this one does not contain PII
    ['codex_cytokit', 'CODEX [Cytokit + SPRM]', False, False]

    >>> name_desc_vis_pii('salmon_rnaseq_bulk')
    ['salmon_rnaseq_bulk', 'Bulk RNA-seq [Salmon]', False, False]

    >>> name_desc_vis_pii('scRNA-Seq-10x')  # This one does contain PII
    ['scRNA-Seq-10x', 'scRNA-seq (10x Genomics)', False, True]

    >>> name_desc_vis_pii(['PAS', 'Image Pyramid'])  # complex alt name
    ['PAS_pyramid', 'PAS Stained Microscopy [Image Pyramid]', True, False]

    >>> name_desc_vis_pii(['Image Pyramid', 'PAS'])  # complex alt name
    ['PAS_pyramid', 'PAS Stained Microscopy [Image Pyramid]', True, False]

    >>> name_desc_vis_pii('image_pyramid')  # simple vis-only alt name
    ['image_pyramid', 'Image Pyramid', True, False]

    >>> name_desc_vis_pii(['Image Pyramid', 'foo'])  # invalid complex alt name
    Traceback (most recent call last):
    ...
    RuntimeError: No such assay_type ['Image Pyramid', 'foo'], even as alternate name

    >>> name_desc_vis_pii('salmon_rnaseq_10x_sn')  # simple valid alt name
    ['salmon_sn_rnaseq_10x', 'snRNA-seq [Salmon]', False, False]
```

src/libs/assay_type.py
```diff
@@ -103,6 +150,21 @@ def iter_names(cls, primary: BoolOrNone = None) -> Iterable[str]:
...
+        >>> all(elt in all_names for elt in primary_names)
+        True
+        >>> any(elt in nonprimary_names for elt in primary_names)
+        False
+        >>> all(elt in all_names for elt in nonprimary_names)
+        True
+        >>> any(elt in primary_names for elt in nonprimary_names)
+        False
+        >>> len(all_names) == len(primary_names) + len(nonprimary_names)
+        True
```
mccalluc, line 166:
```suggestion
        >>> assert all(elt in all_names for elt in primary_names)
        >>> assert not any(elt in nonprimary_names for elt in primary_names)
        >>> assert all(elt in all_names for elt in nonprimary_names)
        >>> assert not any(elt in primary_names for elt in nonprimary_names)
        >>> assert len(all_names) == len(primary_names) + len(nonprimary_names)
```
Your call.

src/libs/test.sh
```diff
@@ -11,6 +11,5 @@ reldir=`realpath --relative-to $topdir .` \
 
 start libs/assay_type
 CMD="python ./assay_type.py"
-diff ${topdir}/examples/libs/assay_type_out.txt <( eval $CMD ) \
-  || die "Try: python ${reldir}/assay_type.py > examples/libs/assay_type_out.txt"
-end libs/assay_type
\ No newline at end of file
+eval $CMD || die "The doctest entries in ${reldir}/assay_type.py need to be updated"
```
mccalluc, line 14:
```suggestion
python ./assay_type.py
```
I think the failure message from doctest is sufficient: Not sure if wrapping it adds value.

mccalluc (approved):
Noted some places where it could be more concise, but these are style tweaks: Your call.

src/libs/assay_type.py
jswelling: Ooh, I didn't know you could put an assert in a doctest.  Yes, this is better.

src/libs/test.sh
jswelling: I'm going to claim that it does add value- I know it does for me, because I'm not used to doctest, and I think it would for Bill.  So I'll leave this one as it is.

jswelling:
This closes https://github.com/hubmapconsortium/ingest-pipeline/issues/317

yuanzhou (approved):

shirey (approved):

mccalluc:
Is the intended use of `contains_pii` to set the visibility of datasets at creation time? There is no call for the portal to use this information at run time in the same way that it might use `vis_only`, I hope?

mruffalo (approved):

jswelling:
@mccalluc as far as I know there is no change to the Portal proposed in response to the contains_pii attribute.  My understanding is that the entity database infrastructure will check this attribute for consistency with its current contains-human-sequence-data flag, and that the Portal bases visibility decisions on that flag, so the Portal should not need to change its behavior.

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -2,18 +2,22 @@ AF:
...
+    vitessce-hints: ['pyramid']
 
 ATACseq-bulk:
     description: Bulk ATAC-seq
     alt-names: []
     primary: true
+    contains-pii: true
     vitessce-hints: []
 
 bulk_atacseq:
```
pdblood, line 23: This is a bad key name, as it represents derived data but it is nearly identical to the raw that that DOES contain PII. The key should have some indication that this is derived data after processing.

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -169,24 +202,25 @@ salmon_rnaseq_bulk:
...
+    contains-pii: true
     vitessce-hints: []
 
 sc_atac_seq_snare_lab:
-#SNAREseq_lab:
     description: snATAC-seq (SNARE-seq2) [Lab Processed]
-#    alt-names: [sc_atac_seq_snare_lab]
     alt-names: []
     primary: false
     vitessce-hints: []
```
pdblood, line 226: @jswelling why don't the lab processed datasets have a `contains-pii` property?

pdblood: same goes for the one below...

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -201,20 +235,21 @@ salmon_rnaseq_snareseq:
     description: snRNA-seq (SNARE-seq2) [Salmon]
     alt-names: []
     primary: false
+    contains-pii: false
     vitessce-hints: ['is_sc', 'rna']
 
 sc_atac_seq_snare:
-#SNAREseq_snapATAC:
```
pdblood, line 207: I don't like this key for the same reason: no indication in the key that it is not raw atac-seq data, which should be protected.

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -223,54 +258,64 @@ salmon_rnaseq_10x:
...
 
 sciATACseq:
     description: sciATAC-seq
     alt-names: []
     primary: true
+    contains-pii: true
     vitessce-hints: []
 
 sc_atac_seq_sci:
     description: sciATAC-seq [SnapATAC]
```
pdblood, line 272: Same complaint as above about this key name not distinguishing from raw protected data...

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -223,54 +258,64 @@ salmon_rnaseq_10x:
...
+    contains-pii: false
 
 snATACseq:
     description: snATAC-seq
     alt-names: []
     primary: true
+    contains-pii: true
     vitessce-hints: []
 
 sn_atac_seq:
```
pdblood, line 321: Same complaint about key name as above...

pdblood (commented):
@jswelling: I think all the PII assignments are correct. The only one I would want to double check out of abundance of caution is DART-FISH... I have a few other comments that you can choose to address now or later.

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -22,72 +26,86 @@ bulk_atacseq:
...
     alt-names: []
     primary: true
     vitessce-hints: []
+    contains-pii: false
 
 image_pyramid:
     description: Image Pyramid
     alt-names: []
     primary: false
+    vis-only: true
```
ropelews, line 63: Why do some of these contain vis-only field and others do not? assuming types marked vis-only=true are ONLY used for visualization.

ropelews (approved):
I think that these are all correctly tagged for pii.

src/search-schema/data/definitions/enums/assay_types.yaml
jswelling: The lab processed datasets don't have it because I wasn't sure if we could trust them to *not* contain PII, so I left it to default to 'does contain PII'.  If we trust them to not include PII in the lab processed results, or if we do plan to actively scan for PII, we could set 'contains-pii' to False for the lab data.  Shall I do so?

src/search-schema/data/definitions/enums/assay_types.yaml
jswelling: The 'primary: true' flag is supposed to distinguish primary datasets.  I agree that a lot of the enums (like sc_atac_seq_snare in this case) are poorly chosen.  Once we have this mechanism in place we can relegate those to being 'alt-names', with better enums for the primary keys.

src/search-schema/data/definitions/enums/assay_types.yaml
jswelling: That one is primary: true, meaning that it is raw data, and contains-pii: true, meaning that it does need to be protected.

src/search-schema/data/definitions/enums/assay_types.yaml
jswelling: That's correct @ropelews .  The vis-ony key defaults to false, just as the contains-pii key defaults to true.

src/search-schema/data/definitions/enums/assay_types.yaml
pdblood: There is a lab processed seqfish that **does** have the explicit `contains-pii: false` label, so I was confused that some lab processed data had it and some didn't. I think we can set these lab processed data to false, and then implement our checking to be extra careful for future data sets. For the lab processed data we have, I think I already verified with them that there was no protected data.

src/search-schema/data/definitions/enums/assay_types.yaml
pdblood: OK, sounds good.

src/search-schema/data/definitions/enums/assay_types.yaml
pdblood: I'm referring to `sc_atac_seq_sci` which is `primary:false`. But, I agree that we can fix these keys later.

jswelling:
I will verify that DART-FISH does not contain PII by looking at a sample, and I will add contains-pii:false to the lab-derived data types.  After that, I will merge.

jswelling:
Our two current DART-FISH samples contain no fastq or BAM files.

PR #332, opened 2021-06-08
Index lab_dataset_id

yuanzhou:
Corresponding to https://github.com/hubmapconsortium/entity-api/pull/187


yuanzhou:
Will need to reindex to get this new field into ES. The ingest-ui uses the result from search-api to display the table view.

PR #341, opened 2021-06-30
Mccalluc/list fields

mccalluc:
This generates the attached report. I'm not sure where's the best place for helper scripts... I might move it down to my directory if you don't want it littering the top level.

[fields.txt](https://github.com/hubmapconsortium/search-api/files/6737623/fields.txt)

- Can you point me at the API documentation that defines each fixed field?
- The index blew up because of the new normalized names: Can you point me at the mapping from old to new, so we can start accommodating that in portal-ui?


yuanzhou:
@mccalluc to answer your questions:

- this yaml file defines all the properties of each Entity type: https://github.com/hubmapconsortium/entity-api/blob/test-release/src/schema/provenance_schema.yaml but not all the defined properties will be exposed via entity-api. And this API documentation (based on the schema yaml) is probably a better version : https://github.com/hubmapconsortium/entity-api/blob/test-release/entity-api-spec.yaml#L198
- The table below shows all the normalized fields (left column) with duplicate data and their current non-normalized fields:

| Neo4j field (normalized) | Existing ES field (to be deprecated) |
| ----------- | ----------------- |
| hubmap_id |  display_doi |
| submission_id | hubmap_display_id |
| (Dataset) ingest_metadata | metadata |
| image_file_metadata | portal_uploaded_image_files |
| (Donor) label | lab_name |
| created_timestamp | create_timestamp |

yuanzhou:
>I'm not sure where's the best place for helper scripts... I might move it down to my directory if you don't want it littering the top level.

At least it needs to be under the `src` directory to be installed and used during the docker build. Moving it down to your directory is probably cleaner. I don't have a strong opinion.

mccalluc:
> At least it needs to be under the src directory to be installed and used during the docker build. 

This is just a tool for developers to run locally, to understand the structure of the index: It is not used by any server-side code: Its one dependency should not be part of the docker build.

> Moving it down to your directory is probably cleaner.

Done. Let me know if anything else is needed to get this merged.

yuanzhou (approved):

PR #343, opened 2021-07-13
Mccalluc/init whitelist

mccalluc:
This is part of the effort to migrate field names (https://github.com/hubmapconsortium/portal-ui/issues/1979). I've cleaned up the portal-ui code for all the fields, except `metadata`: That string gets used so many places, that I don't think I could be sufficiently confident that I had gotten all of the instances (and only the right instances), to tell you to pull the trigger and drop all the old fields from the index. What I'm proposing is a whitelist mechanism that would drop unexpected fields before any other transformations are made. Next steps:

- Review this PR: Do you think this approach will work, or do you think you will need to add fields beyond those that have been wildcarded? Does a flat structure work, or should repeated sets of subfields be defined only once?
- If it can be merged, I would then make a follow-up that would use this whitelist at index time. It could be deployed to production.
- I would then make another follow-up PR that would remove `metadata.*` from the whitelist. This could be deployed to dev/test, and then we could confirm that the portal doesn't depend on any of the `metadata.*` fields.
- That could then be deployed to production.
- Finally, you could remove the old versions of fields from your side: Because we've already seen the effect of removing them at indexing time, there should be no surprises.

While this is feasible, just typing it out, I realize this is a lot of steps. I've assumed that dropping the old versions of fields is an all-or-nothing, but that might be pessimistic... If you could drop them from only the dev or test instance, and then give us time to update the portal-ui code, that could also work.

Another possibility for you would be to drop all the fields, except `metadata`: The rest will be fixed in the portal with today's release. ... But if it's just simpler to do all the fields at once, then ignore that suggestion.


yuanzhou:
@mccalluc maybe things can become simpler if we look at them this way. Currently, we are using the following mapping during indexing time to rename the neo4j filed to the `es_name`:
````
"ingest_metadata": {
  "es_name": "metadata"
},
"metadata": {
  "es_name": "metadata"
},
"specimen_metadata": {
  "es_name": "metadata"
}
````
Based on my findings:
- there's no such `specimen_metadata` field exists currently
- `ingest_metadata` only exists in Dataset
- `metadata` exists only in Donor or Sample, but contains very different sub fields

What I'm proposing is to migrate based on the `entity_type` (loop in @shirey to confirm):
- Dataset: change `metadata` to `ingest_metadata` 
- Donor: keep using `metadata`
- Sample: keep using `metadata`

shirey:
@mccalluc @yuanzhou If you want to change this field name it should be done throughout the Elasticsearch indices and Neo4j.  Id prefer to not change it in one index.  My goal, @yuanzhou , is to make that mapping file that you mention go away as the field names should be common throughout (no mapping needed).
 
In general why can't we use `metadata` across all entity types- why do we need a special name for Datasets?

For `datasets.metadata` I would like to unwind the layers there at some point because there are nested values that don't make sense- I've talked to Joel about this several times, but it is never high enough priority over publishing data to get it done.

shirey (approved):

mccalluc:
> In general why can't we use metadata across all entity types- why do we need a special name for Datasets?

@shirey -- Are you suggesting that the change from `metadata` to `ingest_metadata` doesn't need to happen? I'd be happy with that, and if that's true, then [portal-ui #1979](https://github.com/hubmapconsortium/portal-ui/issues/1979#issuecomment-878535994) could be considered complete.

yuanzhou:
@shirey I'm fine with using `metadata` instead of `ingest_metadata` for Dataset. But we'll also need to change several places in ingest-api and entity-api for renaming to be consistent and not breaking anything else.

mccalluc:
My understanding from Bill at standup is that the status quo is fine for now, but that making `metadata` more consistent is still something he wants to do, but Joel will need to be involved, and no portal work is required right now. Closing, but if I've misunderstood the plan, I apologize.

yuanzhou:
@mccalluc just to confirm, other than this `metadata` filed, the portal-ui has migrated the rest of the normalized fields?

mccalluc:
@yuanzhou -- I believe so, but I'd appreciate it if the index without the duplicated fields could be deployed to dev or test so we can preview it before production.

PR #347, opened 2021-07-15
Reapply the old changes to `dev-integrate`

mccalluc:
This replaces 
- #346
- #320

I just mechanically applied the work that was done earlier -- I haven't taken the time to remind myself of the problem we're trying to solve, or where the best place to make a fix is, given branch structure of this repo -- If I should take the time to understand the big picture, let me know.

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/__init__.py
+++ b/src/elasticsearch/addl_index_transformations/portal/__init__.py
@@ -66,7 +66,7 @@ def transform(doc, batch_id='unspecified'):
     ...        'created_by_user_displayname': 'daniel Cotter'
     ...    }],
     ...    'data_access_level': 'consortium',
-    ...    'data_types': ['codex_cytokit', 'seqFish'],
+    ...    'data_types': ['salmon_rnaseq_10x_sn'],
     ...    'descendants': [{'entity_type': 'Sample or Dataset'}],
     ...    'donor': {
     ...        "metadata": {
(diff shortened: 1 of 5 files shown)


jswelling (changes requested):
One specific problem breaking Travis is that an example in addl_index_transformations/portal/__init__.py is in conflict with an assertion in translate.py .  I think the simplest fix is just to remove the assertion.  No real-world example violates the assertion, but the truth of the assertion is not necessary to the logic.

jswelling:
@mccalluc I thought I'd leave the question of how to deal with the conflict between the test case and the assertion to you.  @yuanzhou a separate question: what is the mechanism that keeps the copies of files under the docker directory consistent with those in the main src tree?  I noticed that they are discordant in the repo.  The assay_types.yaml file differs, for example.

yuanzhou:
@jswelling the answer to your question:
>what is the mechanism that keeps the copies of files under the docker directory consistent with those in the main src tree? I noticed that they are discordant in the repo. The assay_types.yaml file differs, for example.

On DEV, the copies of src files under the docker directory are only used to build the image. And once the container is started the original src files are being mounted from host (the github version) to the container. And this is only for DEV due to frequent code changes.

On TEST/STAGE/PROD, the copies of src files under the docker directory are used to build the image. There's no src mount to the container except the configuration file. Thus whatever files used to build the image are the ones running in the container, regardless of new changes to the original src files. We'll have to rebuild a new image to reflect the src changes.

mccalluc:
- @jswelling / @yuanzhou : Tests are passing; Please review.
- @yuanzhou : You will merge when reviews are in? I think that's the way you like to run this repo.
- @ngehlenborg : You are marked as a codeowner on this, because months ago you wanted to be involved in decisions about how assays are named. Do you want to remain a codeowner here? 

I do not have a strong sense of ownership, but nudge me if I should be thinking more about the underlying problem.

ngehlenborg:
@mccalluc You can remove me as code owner.

yuanzhou (approved):

PR #349, opened 2021-07-16
Mods to assay_types.yaml to distingiush IMC2D from IMC3D

jswelling:
Fix up assay type names to distinguish 2D from 3D IMC, with alt-names for backward compatibility

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -79,7 +79,7 @@ IMC3D:
     vitessce-hints: []
 
 IMC2D_pyramid:
-    description: 2D Imaging Mass Cytometry [Image Pyramid]
+    description: Imaging Mass Cytometry (2D) [Image Pyramid]
     alt-names:
       - 'IMC_pyramid'
       - ['IMC', 'Image Pyramid']
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -65,7 +65,7 @@ image_pyramid:
     vitessce-hints: ['is_image', 'pyramid']
     
 IMC2D:
-    description: 2D Imaging Mass Cytometry
+    description: Imaging Mass Cytometry (2D)
     alt-names: ['IMC', '2D-IMC']
     primary: true
     contains-pii: false
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -64,23 +64,41 @@ image_pyramid:
     contains-pii: false
     vitessce-hints: ['is_image', 'pyramid']
     
-IMC:
-    description: Imaging Mass Cytometry
-    alt-names: []
+IMC2D:
+    description: 2D Imaging Mass Cytometry
+    alt-names: ['IMC', '2D-IMC']
     primary: true
     contains-pii: false
     vitessce-hints: []
 
-3D-IMC:
+IMC3D:
     description: 3D Imaging Mass Cytometry
-    alt-names: []
+    alt-names: ['3D-IMC']
     primary: true
     contains-pii: false
     vitessce-hints: []
 
-IMC_pyramid:
-    description: Imaging Mass Cytometry [Image Pyramid]
-    alt-names: []
+IMC2D_pyramid:
+    description: 2D Imaging Mass Cytometry [Image Pyramid]
+    alt-names:
+      - 'IMC_pyramid'
+      - ['IMC', 'Image Pyramid']
+      - ['Image Pyramid', 'IMC']
+      - ['IMC2D', 'Image Pyramid']
+      - ['Image Pyramid', 'IMC2D']
+    primary: false
+    vis-only: true
+    contains-pii: false
+    vitessce-hints: ['pyramid']
+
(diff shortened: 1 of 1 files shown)


src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -64,23 +64,41 @@ image_pyramid:
     contains-pii: false
     vitessce-hints: ['is_image', 'pyramid']
     
-IMC:
-    description: Imaging Mass Cytometry
-    alt-names: []
+IMC2D:
+    description: 2D Imaging Mass Cytometry
```
ngehlenborg, line 68:
```suggestion
    description: Imaging Mass Cytometry (2D)
```

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -64,23 +64,41 @@ image_pyramid:
...
+    alt-names: ['IMC', '2D-IMC']
```
ngehlenborg, line 69:
```suggestion
    alt-names: ['IMC', 'IMC-2D']
```

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -64,23 +64,41 @@ image_pyramid:
...
     primary: true
     contains-pii: false
     vitessce-hints: []
 
-3D-IMC:
+IMC3D:
     description: 3D Imaging Mass Cytometry
-    alt-names: []
+    alt-names: ['3D-IMC']
```
ngehlenborg, line 76:
```suggestion
    alt-names: ['IMC-3D']
```

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -64,23 +64,41 @@ image_pyramid:
...
     primary: true
     contains-pii: false
     vitessce-hints: []
 
-IMC_pyramid:
-    description: Imaging Mass Cytometry [Image Pyramid]
-    alt-names: []
+IMC2D_pyramid:
+    description: 2D Imaging Mass Cytometry [Image Pyramid]
```
ngehlenborg, line 82:
```suggestion
    description: Imaging Mass Cytometry (2D) [Image Pyramid]
```

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -64,23 +64,41 @@ image_pyramid:
...
+      - ['Image Pyramid', 'IMC']
+      - ['IMC2D', 'Image Pyramid']
+      - ['Image Pyramid', 'IMC2D']
+    primary: false
+    vis-only: true
+    contains-pii: false
+    vitessce-hints: ['pyramid']
+
+IMC3D_pyramid:
+    description: 3D Imaging Mass Cytometry [Image Pyramid]
```
ngehlenborg, line 95:
```suggestion
    description: Imaging Mass Cytometry (3D) [Image Pyramid]
```

src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -64,23 +64,41 @@ image_pyramid:
...
+    alt-names:
```
ngehlenborg, line 96: Where are these alternatives being used? If they show up in user facing parts of the portal, they should also be changed to use "3D" as the postfix rather than prefix wherever possible.

ngehlenborg (changes requested):
Since we use lexicographic order for names and descriptions in several places in the portal, it would make sense to use the subtype (2D, 3D) as a postfix rather than as a prefix so that both show up next to each other.

src/search-schema/data/definitions/enums/assay_types.yaml
jswelling: Concur- I just used the other order because it was more common in our Slack conversations.  I agree your order is preferable.

src/search-schema/data/definitions/enums/assay_types.yaml
jswelling: We can't follow the same policy with alt-names because they are intended to capture and correct existing database entries, replacing the old value '3D-IMC' with the new value IMC3D (with the description we agree on).  3D-IMC is the value that must be replaced so it needs to be the alt-name.

src/search-schema/data/definitions/enums/assay_types.yaml
jswelling: Right- they should never show up in user-facing places.  The alt-names are all deprecated, but examples of metadata files using those names are still in the system.  The alt-name entry allows the deprecated name to be mapped to the correct name before the user sees it.

jswelling: The two element alt-names like ['IMC', 'Image Pyramid'] are meant to match database entries which list those two types, mapping them to the correct description "3D Imaging Mass Cytometry [Image Pyramid]".  The fact that we have image pyramid datasets with the single assay type 'Image Pyramid' and the pair of assay types is an artifact of the fact that 'vis lifting' came along late in the game.

src/search-schema/data/definitions/enums/assay_types.yaml
ngehlenborg: Ok, then let's just do it for those where the change can be made, i.e. not `alt-names`.

ngehlenborg (approved):

PR #352, opened 2021-07-26
Mccalluc/re root still no translation

yuanzhou:
Following merged PR #347 but for `test-release` as the base branch

Will be released along with with https://github.com/hubmapconsortium/commons/pull/56 after being tested on TEST first.


jswelling (approved):
Verified to work on DEV.

PR #354, opened 2021-07-30
Welling/distinguish imc 3d from 2d

yuanzhou:
Following the merged PR #349 but for TEST. FYI @jswelling

Will be released along with with https://github.com/hubmapconsortium/commons/pull/56 after being tested on TEST first.


jswelling (approved):
Confirmed to work on DEV

PR #361, opened 2021-08-20
Fix IMC 3D description fields that were not properly updated in #349

ngehlenborg:

PR #363, opened 2021-08-23
Move "3D" to end for consistency

mccalluc:
Replaces #361: That was going straight into `master`, which I don't think we want to do.


jswelling (approved):
Fine with me.

yuanzhou (approved):

PR #374, opened 2021-09-15
fixed issue for public index access

davism84:


yuanzhou:
Will have another PR to address the issue shortly.

PR #377, opened 2021-09-16
remove deprecated lymph node codes

shirey:


yuanzhou (approved):
I also checked neo4j and we don't have any of those deprecated codes

PR #378, opened 2021-09-21
Use new JSON-LD for anatomy

mccalluc:
### Fix test failures

@yuanzhou / @shirey : I think this should address tests that have been failing since [Bill's commit last week](https://github.com/hubmapconsortium/search-api/commit/762f5aee70b7d59b98194617a57ec67440b80d49) that removed `LY01` etc., but I'm not sure, since CI still isn't running.

I think we're on the same page that moving forward, tests need to be passing before merging to `test-release`? In particular, this means that not even configuration changes should be made directly to `test-release` without PR: Even a configuration change can cause tests to fail. I also think that we're on that same page that it is the responsibility of the Pittsburgh team to keep the CI itself running, whichever service you want to use.

### Upgrade partonomy

@bherr2 : This upgrades the partonomy: It looks like there is good stuff in there, but I think there are still problems. The [top level of the tree](https://github.com/hubmapconsortium/search-api/pull/378/files#diff-55826f44f8623673cb436d1272f9e5a399197923412f52d750465f5d13b87888R76-R81) seems weird:
```
     ...
     'retromandibular vein',
     'skin',
     'spleen',
     'spleen',
     'spleen',
     'telencephalic ventricle',
     ....
```
- Why is `spleen` repeated?
- What's up with `retromandibular vein` and `telencephalic ventricle`? These don't seem like they belong at the top level.

Going forward, as there are new partonomy releases, I think it would make sense for Bruce to start PRs in this repo, and just loop me in as a reviewer if needed.

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/__init__.py
+++ b/src/elasticsearch/addl_index_transformations/portal/__init__.py
@@ -95,9 +95,8 @@ def transform(doc, batch_id='unspecified'):
     >>> del transformed['mapper_metadata']
     >>> pprint(transformed)
     {'anatomy_0': 'body',
-     'anatomy_1': 'abdominal cavity',
-     'anatomy_2': 'colon',
-     'anatomy_3': 'transverse colon',
+     'anatomy_1': 'large intestine',
+     'anatomy_2': 'transverse colon',
      'ancestor_counts': {'entity_type': {}},
      'ancestor_ids': ['1234', '5678'],
      'ancestors': [{'created_by_user_displayname': 'Daniel Cotter',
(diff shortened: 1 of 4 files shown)


bherr2 (approved):

shirey (approved):
Thanks for fixing.

yuanzhou:
Thanks @mccalluc. And we are on the same page:
>I think we're on the same page that moving forward, tests need to be passing before merging to test-release? In particular, this means that not even configuration changes should be made directly to test-release without PR: Even a configuration change can cause tests to fail. I also think that we're on that same page that it is the responsibility of the Pittsburgh team to keep the CI itself running, whichever service you want to use.

I've tested the Travis CI after the payment issue addressed, also updated Github settings of this repo to require status checks to pass before merging.

mccalluc:
@yuanzhou -- great; In this repo, you'll do the merge? Thanks!

yuanzhou (approved):

yuanzhou:
@mccalluc yes, I'll take care of the merge and deployment.

PR #384, opened 2021-09-24
pull out partonomy version number

mccalluc:
Make it easier to remember how to upgrade the partonomy by pulling out to a file by itself. No objection if there's some other configuration file that would be a better place for this to live.


yuanzhou (approved):
@mccalluc will this change require a full reindex?

mccalluc:
> @mccalluc will this change require a full reindex?

This by itself changes nothing about the output; It just surfaces the version number to make it easier for @bherr2 to update.

(I don't think I've seen the result of the reindexing with 1.6.0, but that was an earlier PR.)

bherr2 (approved):

PR #387, opened 2021-09-30
Add support for partonomy placement using organ codes

bherr2:
Added support for partonomy placement using organ codes in addition to ccf_annotations. This means that all docs can be placed in the partonomy/anatomical structure search. Ideally the IRIs come from somewhere else, but right now I'm placing them as a lookup in the add_partonomy.py file.


yuanzhou:
@bherr2 did you create your branch off “test-release”? This is the default branch that also contains the recent changes. We only use master branch for release purpose.

bherr2:
> @bherr2 did you create your branch off “test-release”? This is the default branch that also contains the recent changes. We only use master branch for release purpose.

Oh ok, I didn't know. Anything I need to change? Looks like you already switched the branch for me.

yuanzhou:
@bherr2 yes I’ve changed the base branch. And the CI checks all passed after you updated your branch. Thanks. 

@mccalluc i don’t know much about the changes. If you think they look good, I’ll get this PR merged and redeployed on TEST.

src/elasticsearch/addl_index_transformations/portal/add_partonomy.py
```diff
@@ -2,6 +2,30 @@
...
+    'OT': 'http://purl.obolibrary.org/obo/UBERON_0013702'
+}
+
+
+def get_organ_iri(doc):
+    if 'origin_sample' in doc and 'organ' in doc['origin_sample']:
+        organ_code = doc['origin_sample']['organ']
+        if organ_code in ORGAN_IRI:
+            return ORGAN_IRI[organ_code]
+    return None
```
mccalluc, line 27: We already have code and configuration for mapping Elasticsearch organ terms to IRIs in [`portal-ui/organ-utils`](https://github.com/hubmapconsortium/portal-ui/blob/main/organ-utils/make_organ_dir.py#L192-L200) -- I don't think this logic should be repeated across the codebase.

src/elasticsearch/addl_index_transformations/portal/add_partonomy.py
```diff
@@ -13,23 +37,31 @@ def add_partonomy(doc):
...
+    ...     'origin_sample': {
+    ...         'organ': 'LI'
+    ...     },
     ...     'rui_location': dumps(rui_location)
     ... }
     >>> add_partonomy(doc)
     >>> del doc['rui_location']
+    >>> del doc['origin_sample']
     >>> doc
     {'anatomy_0': 'body', 'anatomy_1': 'large intestine', 'anatomy_2': 'transverse colon'}
```
mccalluc, line 49: These are two separate execution paths, so we need two separate tests, at least.

src/elasticsearch/addl_index_transformations/portal/add_partonomy.py
```diff
@@ -13,23 +37,31 @@ def add_partonomy(doc):
...
+        annotations.append(organ_iri)
+
+    if 'rui_location' in doc:
+        rui_location = loads(doc['rui_location'])
 
-    if 'ccf_annotations' not in rui_location:
-        return
-    ccf_annotations = rui_location['ccf_annotations']
+        if 'ccf_annotations' in rui_location:
+            annotations += rui_location['ccf_annotations']
```
mccalluc, line 62:
```suggestion
        annotations += rui_location.get('ccf_annotations', [])
```

mccalluc (changes requested):
Thanks -- We will build on this. Right now, I don't want to introduce a duplicate mapping between Elasticsearch and the IRIs. I think a submodule would work, but want to confirm with @shirey .

src/elasticsearch/addl_index_transformations/portal/add_partonomy.py
bherr2: Oops, commit suggestion didn't work right. Anyway, I'll let you take over from here.

mccalluc:
I've put getting a single-source-of-truth for organs on the agenda for tomorrow.

mccalluc:
October 1, 2021 Jonathan asserted that the Ontology API is the right place to capture this information, and it will be in place soon, and will be accessed via hubmap-commons.

bherr2:
Oh jesus, are you really closing this PR because of that? Just accept that not everything can be perfect 100% of the time and add this useful bit of code.

mccalluc:
Waiting on mapping to be provided in hubmap-commons: https://github.com/hubmapconsortium/commons/issues/61

mccalluc:
Talked with Nils: He wants to get this in without waiting for the backing ontology work. I'll make a commit to add a test case that distinguishes the two paths.

mccalluc (approved):

mccalluc:
- Added the IRIs to `organ_types.yaml`, rather than starting another list that will need to be merged.
- If there are multiple partonomy nodes, include all paths in the facets we output. In the fake test examples, these don't make much sense, but hopefully in the real world, the IRI from the two letter code would already be in what we get from the RUI-JSON. (ES doesn't care whether a single value or a single-valued-list is given: They index the same.)
- Test cases for the different combinations.

@yuanzhou : In your court.

yuanzhou (approved):

yuanzhou:
@bherr2 @mccalluc FYI, I've redeployed the changes on DEV and TEST, will also do a reindexing.

PR #406, opened 2021-10-26
Mods to add MS and LC-MS bottom-up and top-down, and deletions

jswelling:
This PR adds MS and LC-MS assay types, with and without Bottom-Up and Top-Down variants.  It deletes Targeted-Shotgun-LC-MS, LC-MS-untargeted, and TMT-LC-MS.  In addition to merging, it will require 65 existing datasets with the deleted types to be edited.  An update to the workflow map in ingest-pipeline will also be needed.


yuanzhou:
Is this PR still relevant? Please close if not.

yuanzhou:
Close this PR for now, reopen if needed.

PR #408, opened 2021-10-27
Mccalluc/mapped ext consortium

mccalluc:
Towards https://github.com/hubmapconsortium/portal-ui/issues/1406.

Screen Shot 2021-10-27 at 1 46 04 PM

@ngehlenborg : Can you confirm that 
- you are happy with the current facet string for HCA?
- but on the details page, in the header, you would prefer a string without `EXT`? [design](https://github.com/hubmapconsortium/portal-ui/issues/1297#issuecomment-740089734) 

@shirey 
- Assuming that Nils is happy with it, can you confirm that all future external datasets will also include `EXT - ` in their `group_name`?

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/__init__.py
+++ b/src/elasticsearch/addl_index_transformations/portal/__init__.py
@@ -56,6 +56,7 @@ def transform(doc, batch_id='unspecified'):
     >>> transformed = transform({
     ...    'entity_type': 'dataset',
     ...    'status': 'New',
+    ...    'group_name': 'EXT - Outside HuBMAP',
     ...    'origin_sample': {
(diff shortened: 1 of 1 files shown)


yuanzhou:
@mccalluc are there any status change on this PR?

ngehlenborg (approved):
I am happy with the current facet string for HCA and on the details page, in the header, I would prefer a string without EXT!

shirey (approved):
I originally added the "EXT" to mimic the "TMC/TTD/etc.." that we have for other groups and thought it should be treated the same for these external groups as we treat TMC/TTD/etc... for the internal Consortium groups.

yuanzhou (approved):

PR #409, opened 2021-10-27
Add support for scRNAseq-10xGenomics, -v2, and -v3

jswelling:
This PR adds support for the new assay type names scRNAseq-10xGenomics, scRNAseq-10xGenomics-v2, and scRNAseq-10xGenomics-v3 .  -v3 is assumed to be the common case, and thus replaces the old name scRNA-Seq-10x.  The version without the suffix is assumed to be a redundant name for that same assay.  The -v2 variant is assumed to be a new assay type.  

The derived dataset type produced from all of these by the salmon_rnaseq pipeline remains salmon-rnaseq_10x, so note that these two different primary assay types give rise to the same secondary dataset type.

In addition to this PR, these changes will require a PR against ingest-pipeline to update the workflow map and to set appropriate flags for processing -v2 data as having chromium-v2 tags.

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -394,16 +394,9 @@ sn_atac_seq:
     contains-pii: false
     vitessce-hints: ['is_sc', 'atac']
 
-snRNAseq-v2:
-    description: 'snRNA-seq (v2)'
+snRNAseq:
+    description: snRNA-seq
     alt-names: []
-    primary: true
-    contains-pii: true
-    vitessce-hints: []
-
-snRNAseq-v3:
-    description: 'snRNA-seq (v3)'
-    alt-names: ['snRNAseq']
     primary: true
     contains-pii: true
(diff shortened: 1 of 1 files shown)


jswelling:
@mruffalo but we do have datasets with assay type snRNA-seq which have been on hold because they are apparently chromium-v2: 488f364142c308a9692e0b529f6697dd and 4bef8fa6eab2d3eb8734bf418c0634ef . My understanding has been that 10X is necessarily sc rather than sn.  Do we actually need to bring the -v2/-v3 distinction to snRNA-seq, and if so, how should we do so?

mruffalo:
@jswelling we have 10X single-cell and 10X single-nucleus data in the portal now, so I would propose using `snRNAseq-10xGenomics-v[23]` instead of `snRNAseq-v[23]`.

jswelling:
@mruffalo I've separated the snRNAseq changes off into a separate PR https://github.com/hubmapconsortium/search-api/pull/410 .

mruffalo (approved):

PR #417, opened 2021-12-03
Add NanoDESI, NanoPOTS, multiple MS and LC-MS assays

jswelling:
This adds new assays: NanoDESI and its image pyramid, NanoPOTS (with no image pyramid), and the normal, Bottom Up, and Top Down variants of LC-MS and MS.  I believe none of these types contain personally identifying information.  The older equivalent LC-MS assays are *not* removed because we have not yet come to consensus on how to reclassify the existing datasets, so old and new LC-MS versions will exist together in the database until that consensus is reached.  

A block of comments preserving older MALDI positive and negative definitions was removed, since that change has passed testing.

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -184,7 +184,7 @@ NanoDESI:
 
 NanoDESI_pyramid:
   description: NanoDESI [Image Pyramid]
-  alt-names: [["NanoDesi", "Image Pyramid"], ["NanoDesi", "image_pyramid"]]
+  alt-names: [["NanoDESI", "Image Pyramid"], ["NanoDESI", "image_pyramid"]]
   primary: false
   vis-only: true
   contains-pii: false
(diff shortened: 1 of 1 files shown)


mccalluc:
Removing myself from the reviewers list: What really needs to be checked here is the content, and that's not an area where I need to be involved. Also, if responsibility is spread across too many people, everyone may think that someone else is giving it a close reading.

yuanzhou:
@jswelling I looked at the PROD neo4j, currently we still have lots of datasets with the following types:
- MALDI-IMS-neg
- MALDI-IMS-pos

There is no match for 
- ALDI-IMS-neg_pyramid
- MALDI-IMS-pos_pyramid

The tests passed with commenting out these 4 types, how should we handle the existing datasets with the above two types? Also include @shirey to this discussion.

jswelling:
@mccalluc the thing I want to bring to your attention is that this PR introduces the new MS and LC-MS assay types without removing the old types, which include existing datasets ultimately needing reclassification as new types.  I've proposed this so that we can begin to ingest new datasets in the interim.  I am concerned that this is something you will feel strongly about.

yuanzhou:
@jswelling thanks for explaining the details.

yuanzhou (approved):
I'll merge this PR and redeploy on TEST once all other reviewers have approved.

cebriggs7135 (approved):

ngehlenborg (approved):

pdblood (approved):

PR #419, opened 2021-12-07
'sprm', 'anndata', 'multiscale' instead of 'codex'

jswelling:


src/search-schema/data/definitions/enums/assay_types.yaml
```diff
@@ -40,7 +40,7 @@ celldive_deepcell:
     alt-names: []
     primary: false
     contains-pii: false
-    vitessce-hints: ['codex', 'is_image', 'is_tiled']
+    vitessce-hints: ['sprm', 'anndata', 'multiscale', 'is_image', 'is_tiled']
```
ilan-gold, line 43: ok i forgot about `is_tiled` so we can drop `multiscale`

jswelling: Done.

PR #422, opened 2021-12-10
Mccalluc/numbers should be numbers

### LINKED ISSUE #421: Make sure metadata numbers are numbers
For ES sorting to work, fields must be of the correct type... numbers need to be numbers.

Another possibility would be to move this responsibility back to the TSV parser.

But simpler to just try parsing metadata values to numbers, and using the numbers.

mccalluc:
- Fix #421

I don't know that this will do what we want, but I think it's worth trying. The possible downside is fields outside the metadata being treated as numbers, with unexpected consequences. We might also hit a problem if the first document with a given field looks like a number... but later documents reveal it to be a string.

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/__init__.py
+++ b/src/elasticsearch/addl_index_transformations/portal/__init__.py
@@ -219,8 +219,6 @@ def _simple_clean(doc):
                 if v in ['1', 'true', 'True']:
                     metadata[k] = 'TRUE'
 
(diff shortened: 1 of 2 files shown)


john-conroy (approved):
Worth trying, but strings mistaken as numbers will cause the same problems when sorting.

yuanzhou (approved):
Will get it out on TEST to find out.

PR #435, opened 2022-01-24
Add file extension to description

mccalluc:
For file manifest generation, knowing the type of the file may be important to the user. The same information may be available in multiple formats. (ie, a QA report available as text or pdf or html...)

I believe that adding the file extension to the description was what we settled on, but there are other ways this could be done:
- upper vs lower case?
- leading `.` or no `.`?
- maybe skip this if the extension is already in the description... but the consistency may be a good thing?

Or we could take an entirely different approach...
- Elasticsearch [nested fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html) work like sub-documents inside the parent document.
- Or (and I don't think this is the direction we're going) there could be a separate ES index just for the files. [Proposal](https://docs.google.com/document/d/1zZaX64Wx_f0j0lQWw4mA7DHpjA5rQ-CKmjIx6k6UpoE/edit).

@tsliaw -- Confirm that the resulting strings satisfy the requirements of the UI you're working on?

cc @mruffalo -- I think you've been providing the descriptions. Mostly just a heads up about how we intend to use the information you provide.

### DIFF
--- a/src/elasticsearch/addl_index_transformations/portal/__init__.py
+++ b/src/elasticsearch/addl_index_transformations/portal/__init__.py
@@ -80,6 +80,14 @@ def transform(doc, batch_id='unspecified'):
     ...            ]
     ...        }
     ...    },
+    ...    'files': [{
+    ...        "description": "OME-TIFF pyramid file",
+    ...        "edam_term": "EDAM_1.24.format_3727",
+    ...        "is_qa_qc": False,
+    ...        "rel_path": "ometiff-pyramids/stitched/expressions/reg1_stitched_expressions.ome.tif",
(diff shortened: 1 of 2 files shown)


yuanzhou:
I don't know much about the context here, will handle the merge and deployment once the other reviewer approves.

tsliaw:
The mock-up doesn't have the file extension in the description, but it might be a good step forward. 

> upper vs lower case?

Upper case

> leading . or no .?

No leading '.'

> maybe skip this if the extension is already in the description... but the consistency may be a good thing?

Let's have consistency here

tsliaw (approved):

PR #442, opened 2022-02-01
Mccalluc/add portal visualization

mccalluc:
The goal is to move vis-conf build to index-time so that problems are encountered earlier, rather than deferred until render-time. This is the start of alternative to:
- https://github.com/hubmapconsortium/portal-ui/pull/2313

Vis config code is incorporated as a git submodule so that it can be shared between this this repo and portal-ui. As of
- https://github.com/hubmapconsortium/portal-ui/pull/2386

the submodule is being used at render time... once the view conf is available in the index, the plan is to just keep the submodule in portal-ui around for developer use.

Filing as draft until everything works... but early feedback welcome.


mccalluc:
Asking Zhou for help on Slack:
> What is the right way to access the TypeClient at index-time? We are moving away from accessing it in the portal at render-time: In that context we have access to a TYPE_SERVICE_ENDPOINT from the Flask context… what’s the right way to do this in search-api?

mccalluc:
From Joel:
> I believe if you are running in the context of search-api you can probably instantiate libs.assay_type.AssayType(type_name) directly.

mccalluc:
At the meeting this morning, concluded that this is not a viable solution. Closing.

PR #446, opened 2022-02-07
Added .idea to .gitignore and hm_antibody to search-config.yaml.example

ChuckKollar:


yuanzhou (changes requested):
I looked at the above tweak made in indexer.py to skip the self-managed indices like antibody index, there are some 
other places that would also need to be taken care of:
- https://github.com/hubmapconsortium/search-api/blob/hm_antibodies_index/src/elasticsearch/indexer.py#L86
- https://github.com/hubmapconsortium/search-api/blob/hm_antibodies_index/src/elasticsearch/indexer.py#L267
- https://github.com/hubmapconsortium/search-api/blob/hm_antibodies_index/src/elasticsearch/indexer.py#L410
- https://github.com/hubmapconsortium/search-api/blob/hm_antibodies_index/src/elasticsearch/indexer.py#L807-L812

Instead of handling this special case in multiple places, it's better to take care of it in one place by filtering out the ones that are not used by indexer for reindeixing especially. 

I don't see where this top-level config `active: true|false` being used, would it be a good idea if we rename it to something like `reindex_enabled: true|false` and set it to false on antibody index. Then only keep the indices that are `reindex_enabled:true` at https://github.com/hubmapconsortium/search-api/blob/hm_antibodies_index/src/elasticsearch/indexer.py#L54

ChuckKollar (commented):
Here the 'indices' keys that are not marked with 'reindex_enabled': true will not be included for consideration. This is done in the Indices.__init__() method.

ChuckKollar (commented):
Fixed 'true' -> True and indices -> 'indices' and tested.

yuanzhou (approved):

PR #456, opened 2022-03-07
Generalize search branch

maxsibilla:
Modifying architecture to utilize an updated interface design for the translation portion of indexing. Created two subclasses responsible for creating documents for SenNet and HuBMAP. Also created a standalone multiplexer indexer class that is solely responsible for indexing.


yuanzhou:
@maxsibilla there are lots of changes so I'll compose a list of items:
- Can you sync the latest `test-release` branch of HuBMAP search-api repo into your branch? This `git pull` should resolve a few merge conflicts as well.
- It's suggested that we use all-lowercase for Python file/module naming convention. So can you rename `src/Indexer.py` to something like `indexer_base.py` and the modules in `src/translator` with lowercase and underscore as well?
- There's mixed use of Flask within the `src/elacticsearch/indexer.py` https://github.com/sennetconsortium/search-api/blob/generalize-search-branch/src/elasticsearch/indexer.py#L40-L42  and `src/translator/translation_functions.py` https://github.com/sennetconsortium/search-api/blob/generalize-search-branch/src/translator/translation_functions.py#L31-L48 In my opinion, Flask should only be used in `app.py` when handling REST API requests/responses. The indexer and translator functions should be self-contained and unrelated to Flask framrwork.
- The `src/translator/translation_functions.py` also handles the search execution and other REST API related functions, while they should be in a separate package/module since they are not related to the translation per-se. 
- In terms of file structure, I don't know how should we organize the `translator` package and the `elasticsearch` package, should the translator be a sub directory with in the elasticsearch?
- We should have separate `neo4j-to-es-attributes.json` for HuBMAP and SenNet since the entities schema will be different, I don't know if this has been taken into consideration in your generalization.

davism84 (commented):
Unless you are going to use the same transformations, as HubMap, which these only apply to the portal, I don't think there should be any changes the portal/translation code.

maxsibilla:
I am still performing some testing from devops but I have completed the following:

- I have now merged with the latest test-release branch
- I modified the file names to conform to proper naming convention
- I reverted some functions back to app.py and created a translation specific file with shared functions between the sennet and hubmap translators
- Regarding the elasticsearch vs translation file structure I copied the contents of elasticsearch into "/translator/sennet_translation" and modified the files to conform to with the sennet versions of the entities. We can simply create a "hubmap_translation" directory and just delete elasticsearch but we can discuss
- Due to the modifications above there are 2 separate neo4j-to-es-attributes.json files
- I have modified the files you listed to place the __init__ functions at the beginning of the class

yuanzhou:
Will merge and test on DEV

yuanzhou (approved):

PR #486, opened 2022-04-08
Search-api generalization

maxsibilla:
- Generalized the index/reindex process with abstracting the translator to be used by more projects. Handled by new config settings in app.cfg.
- Changes already in place in AWS API Gateway to reflect the two new endpoints `POST /add/<id>` and `PUT /update/<id>` with requiring tokens with at least one data provider associated with. These endpoints accept JSON document to be directly indexed into elasticsearch.


yuanzhou:
Close this PR due to the changes needed for sub-module and docker.

PR #490, opened 2022-04-25
Submodule branch

yuanzhou:
All the generalization works that @maxsibilla made


yuanzhou:
We can't really use Travis here due to the breaking changes/refactoring.

PR #532, opened 2022-07-05
Updating dynamic template to treat any field inside metadata.metadata as a string

yuanzhou:


yuanzhou:
Moving to TEST after DEV test fine.

PR #540, opened 2022-07-15
Derek furst/expose metadata

DerekFurstPitt:
If ingest_metadata or ingest_metadata.files is empty in hubmap_translator.py, `"files": []` is added to the top level "entity"

### DIFF
--- a/src/hubmap_translator.py
+++ b/src/hubmap_translator.py
@@ -663,7 +663,16 @@ class Translator(TranslatorInterface):
                     if 'ingest_metadata' in entity:
                         ingest_metadata = entity['ingest_metadata']
                         if 'files' in ingest_metadata:
-                            entity['files'] = ingest_metadata['files']
+                            if ingest_metadata['files'].strip() == "":
+                                entity['files'] = []
+                            else:
+                                entity['files'] = ingest_metadata['files']
+                        else:
+                            entity['files'] = []
+                    else:
+                        entity['files'] = []
+
+
 
             self.entity_keys_rename(entity)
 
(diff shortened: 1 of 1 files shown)


src/hubmap_translator.py
```diff
@@ -651,7 +663,16 @@ def generate_doc(self, entity, return_type):
                     if 'ingest_metadata' in entity:
                         ingest_metadata = entity['ingest_metadata']
                         if 'files' in ingest_metadata:
-                            entity['files'] = ingest_metadata['files']
+                            if ingest_metadata['files'].strip() == "":
```
yuanzhou, line 666: Is this the assumption that the `Dataset.ingest_metadata.files` field is ALWYAS either a list of dicts or empty string? Otherwise, could it be an empty list `[]` as well?

@DerekFurstPitt this is not a change request, it's more about confirming all the possible cases with Joel.

I also noticed that we have this existing block: https://github.com/hubmapconsortium/search-api/blob/Derek-Furst/expose-metadata/src/hubmap_translator.py#L692-L695 that removes the `metadata.files` after renaming from the `ingest_metadata`. I'm wondering if we should move this block right below your changes and keep using `ingest_metadata` since it's not yet renamed at that point when IF we move it there. Just make it more readable?

DerekFurstPitt: @yuanzhou Updated and repushed

yuanzhou (approved):
@shirey the changes @DerekFurstPitt made looks good now, should we hold off merging them due to the index issue we are trying to resolve?

yuanzhou:
FYI, I had to change the base branch to a different branch then switch back to update this PR. Otherwise it's showing 8 files changed instead of just 1.

yuanzhou (approved):

PR #553, opened 2022-07-25
applied a fix to hubmap_translator.py that caused 'files' to not prop…

yuanzhou:


yuanzhou:
Replaced by PR #555 with more fixes

PR #562, opened 2022-09-07
Mccalluc/combine requirements

mccalluc:
`requirements.txt` can use `-r` internally: This lets us specify the transitive requirements in one place, rather than repeating ourselves. It also insures that there aren't dependency conflicts: Right now the dependency versions aren't actually consistent, and the state of the system relies on the order of installation (fragile) rather than pip's dependency resolution (robust). 
```
$ grep Flask src/requirements.txt | grep -v '#'
Flask==2.1.3
```
conflicts with:
```
grep Flask src/search-adaptor/src/requirements.txt | grep -v '#'
Flask==1.1.2
```
but note that:
```
$ cd src/search-adaptor/; git checkout main; cd -
$ grep Flask src/search-adaptor/src/requirements.txt | grep -v '#'
Flask==2.1.3
```

This PR updates search-adaptor to the current version on `main` to resolve this conflict, but you might want something else.

Suggestions:
- libraries (like search-adaptor) should specify ranges on their dependencies so that when they are used with other libraries the risk of incompatible requirements is minimized.
- applications (like search-api) should record the exact set of requirements that are known to work. There are a number of different solutions for this in the python world, including pip freeze, Pipfile, and requirements.in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ services:
 before_script:
   - sleep 10  # Advised by https://docs.travis-ci.com/user/database-setup/#elasticsearch
 install:
-  - pip install -r src/search-adaptor/src/requirements.txt
   - pip install -r src/requirements.txt
   - pip install -r src/requirements-dev.txt
 script:
--- a/docker/search-api/Dockerfile
+++ b/docker/search-api/Dockerfile
@@ -40,7 +40,7 @@ RUN yum install -y yum-utils && \
     rm /etc/nginx/conf.d/default.conf && \
     mv nginx/nginx.conf /etc/nginx/nginx.conf && \
     rm -rf nginx && \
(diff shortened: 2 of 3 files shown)


yuanzhou:
@mccalluc the search-adaptor is using the same Flask version: https://github.com/dbmi-pitt/search-adaptor/blob/main/src/requirements.txt#L1 when we upgraded to Flask v2 in both the submodule and search-api.
>conflicts with:
>>grep Flask src/search-adaptor/src/requirements.txt | grep -v '#'
Flask==1.1.2

yuanzhou:
Thanks for the suggestion and I've made the corresponding changes in the submodule:
>libraries (like search-adaptor) should specify ranges on their dependencies so that when they are used with other libraries the risk of incompatible requirements is minimized.

yuanzhou (approved):

PR #564, opened 2022-09-07
Mccalluc/reboot schema

mccalluc:
- Previous attempt by Harvard to prescribe document structure never converged with what the document structure actually was.
- Development on HuBMAP has been impeded on multiple occasions by unexpected changes to document structure.
- This PR scraps the previous work, and proposes building a new, useful schema in small increments.
- Approval and merge will indicate that there is a shared commitment to avoiding surprise changes to document structure, reviewing the `validation_errors` after index, and if there are any, resolving them.

cc @ngehlenborg


docker/search-api/.dockerignore
```diff
@@ -9,6 +9,3 @@ nginx/conf.d
 
 # Ignore the flask instance folder
 src/instance
-
-# Ignore search-schema from the submodule
```
yuanzhou, line 13: There's no longer this `src/search-schema` sub directory in the submodule: https://github.com/dbmi-pitt/search-adaptor are you using an older version?

yuanzhou (changes requested):

docker/search-api/.dockerignore
mccalluc: Right: If it's no longer there, we no longer need to dockerignore it. This line doesn't do any harm, but we shouldn't keep vestigial code around either.

yuanzhou: Cool, I'll just delete that line from dockerignore to keep things consistent and updated to avoid future confusion then merge.

yuanzhou (approved):

PR #573, opened 2022-09-23
Derek furst/reapply source origin samples

DerekFurstPitt:
Reimplemented source samples and origin samples in search-api

### DIFF
--- a/src/hubmap_translator.py
+++ b/src/hubmap_translator.py
@@ -644,18 +644,19 @@ class Translator(TranslatorInterface):
             if entity['entity_type'] in ['Sample', 'Dataset']:
                 # Add new properties
                 entity['donor'] = donor
-
-                entity['origin_sample'] = copy.copy(entity) if ('specimen_type' in entity) and (
-                        entity['specimen_type'].lower() == 'organ') and ('organ' in entity) and (
-                                                                       entity['organ'].strip() != '') else None
+                entity['origin_sample'] = copy.copy(entity) if ('specimen_type' in entity) and (entity['specimen_type'].lower() == 'organ') and ('organ' in entity) and (entity['organ'].strip() != '') else None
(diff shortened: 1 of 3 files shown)


src/hubmap_translator.py
```diff
@@ -644,43 +647,58 @@ def generate_doc(self, entity, return_type):
...
                             if parents[0]['entity_type'] == 'Sample':
-                                # entity['source_sample'] = parents[0]
                                 entity['source_sample'] = parents
 
                             e = parents[0]
                         except IndexError:
                             entity['source_sample'] = {}
 
+                    e = entity
+                    entity_sdk_instance = EntitySdk(token=self.token, service_url=self.entity_api_url)
```
yuanzhou, line 686: @DerekFurstPitt we need to back out the use of hubmap_sdk, two reasons:
- there's a dependency conflict. hubmap-sdk 1.0.2 depends on urllib3==1.26.7 but the current search-api uses urllib3==1.26.5. The build failed.
- we should still use the `call_entity_api()` method to retrieve the parent entities, for now, to keep the code consistent with the calls used elsewhere. 

Replacing with the hubmap_sdk can be done separately later, and when we do that we should replace everywhere the old calls being used. For the purpose of this PR, we don't want to introduce new changes that are unrepeated.

yuanzhou (changes requested):

src/hubmap_translator.py
DerekFurstPitt: Updated. I understand. I was just trying to save myself some work down the road when inevitably hubmap_sdk gets applied throughout search-api.

src/requirements.txt
```diff
@@ -1,7 +1,7 @@
 Flask==2.1.3
 urllib3==1.26.5
 jsonschema==3.2.0
-
+hubmap-sdk==1.0.2
```
yuanzhou, line 4: @DerekFurstPitt you'll also need to remove this dependency otherwise the build still fails.

yuanzhou (changes requested):

yuanzhou (approved):
Holding off on the merge, will need to release other changes first. Also need to coordinate with Harvard team to figure out a plan for to use the new fields and deprecate the old fields.

PR #584, opened 2022-10-28
Karlburke/files api

kburke:
search-api endoints to work with Elasticsearch index of Dataset file manifests.  Soon to be integrated with files-api revisions.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -38,4 +38,4 @@ services:
 networks:
   # This is the network created by gateway to enable communicaton between multiple docker-compose projects
   gateway_hubmap:
-    external: true
+    external: false
--- a/src/hubmap_translator.py
+++ b/src/hubmap_translator.py
@@ -219,31 +219,31 @@ class Translator(TranslatorInterface):
 
     def update(self, entity_id, document, index=None):
         if index is not None:
-            self.indexer.index(entity_id, json.dumps(document), index, True)
+            response = self.indexer.index(entity_id, json.dumps(document), index, True)
         else:
             for index in self.indices.keys():
                 public_index = self.INDICES['indices'][index]['public']
                 private_index = self.INDICES['indices'][index]['private']
 
                 if self.is_public(document):
-                    self.indexer.index(entity_id, json.dumps(document), public_index, True)
+                    response = self.indexer.index(entity_id, json.dumps(document), public_index, True)
 
-                self.indexer.index(entity_id, json.dumps(document), private_index, True)
-        return "Updated document successfully"
+                response += self.indexer.index(entity_id, json.dumps(document), private_index, True)
+        return response
 
     def add(self, entity_id, document, index=None):
         if index is not None:
-            self.indexer.index(entity_id, json.dumps(document), index, False)
+            response = self.indexer.index(entity_id, json.dumps(document), index, False)
         else:
             for index in self.indices.keys():
                 public_index = self.INDICES['indices'][index]['public']
                 private_index = self.INDICES['indices'][index]['private']
 
                 if self.is_public(document):
-                    self.indexer.index(entity_id, json.dumps(document), public_index, False)
+                    response = self.indexer.index(entity_id, json.dumps(document), public_index, False)
 
-                self.indexer.index(entity_id, json.dumps(document), private_index, False)
-        return "Added document successfully"
+                response += self.indexer.index(entity_id, json.dumps(document), private_index, False)
+        return response
 
 
     # Collection doesn't actually have this `data_access_level` property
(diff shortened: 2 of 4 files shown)


shirey (approved):
I approve, but defer to @yuanzhou on the changes in .gitmodues and docker-compose.yml, assuming at least .gitmodules will need to change to match the search-adaptor branch that we're using.

docker/docker-compose.yml
```diff
@@ -38,4 +38,4 @@ services:
 networks:
   # This is the network created by gateway to enable communicaton between multiple docker-compose projects
   gateway_hubmap:
-    external: true
```
yuanzhou, line 41: @kburke we have to use the external docker network especially when we deploy on DEV/TEST/STGAE/PROD. Please revert this change.

src/hubmap_translator.py
```diff
@@ -21,7 +21,7 @@
 from translator.tranlation_helper_functions import *
 from translator.translator_interface import TranslatorInterface
 
-logging.basicConfig(format='[%(asctime)s] %(levelname)s in %(module)s: %(message)s', level=logging.DEBUG,
+logging.basicConfig(format='[%(asctime)s] %(levelname)s in %(module)s: %(message)s', level=logging.INFO,
```
yuanzhou, line 24: We also use the DEBUG level for troubleshooting on the indices config for deployment. Please revert.

src/hubmap_translator.py
```diff
@@ -82,6 +82,8 @@ def __init__(self, indices, app_client_id, app_client_secret, token):
         self.token = token
         self.request_headers = self.create_request_headers_for_auth(token)
         self.entity_api_url = self.indices[self.DEFAULT_INDEX_WITHOUT_PREFIX]['document_source_endpoint'].strip('/')
+        UUID_API_URL = 'https://uuid-api.dev.hubmapconsortium.org' #@TODO-un-hard-code, figure out what stuff on above line does/means
```
yuanzhou, line 85: We should add a new config for this uuid-api URL in https://github.com/hubmapconsortium/search-api/blob/karlburke/files-api/src/instance/search-config.yaml.example and the actual location depends on how this uuid-api URL gets called. 

https://github.com/hubmapconsortium/search-api/blob/karlburke/files-api/src/hubmap_translator.py#L872 uses the uuid-api URL, but I can't find how this method is being used.

src/hubmap_translator.py
```diff
@@ -828,17 +857,35 @@ def prepare_dataset(self, dataset_dict):
...
                     if isinstance(dataset_dict['ingest_metadata']['metadata'][key], str):
                         if not dataset_dict['ingest_metadata']['metadata'][key] or re.search(r'^\s+$', dataset_dict['ingest_metadata']['metadata'][key]):
                             del dataset_dict['ingest_metadata']['metadata'][key]
                             logger.info(f"Removed ['ingest_metadata']['metadata']['{key}'] due to empty string value, for Dataset {dataset_dict['uuid']}")
-            
+
         return dataset_dict
 
+    # Retrieve a list of files in a Dataset from the uuid-api service
+    def _get_dataset_files_info(self, dataset_uuid):
```
yuanzhou, line 872: Is this `_get_dataset_files_info()` being used somwhere?

src/instance/search-config.yaml.example
```diff
@@ -41,3 +41,11 @@ indices:
...
     elasticsearch:
       url: <ES URL>
+
+  hm_files:
+    active: true
+    reindex_enabled: false
+    public: hm_files
+    elasticsearch:
+      url: <ES URL>
+      mappings: "hubmap_translation/search-default-config.yaml"
```
yuanzhou, line 51: We probably need a file-api specific config yaml, @maxsibilla what do you think?

yuanzhou (changes requested):

docker/docker-compose.yml
kburke: Setting reverted to true in branch file.

src/hubmap_translator.py
kburke: Logging level reverted to DEBUG in branch file.

src/hubmap_translator.py
kburke: The method using these variables was not being used, per comment on your observation below. Removed in branch file.

yuanzhou (approved):

src/instance/search-config.yaml.example
maxsibilla: @yuanzhou I don't have much experience with the antibodies ES instance so I don't know if the files component breaks that convention. It might be best to go forward with a separate config file just for the reason that it does not have a private and public index and the `main` function in `hubmap_translator.py` might have to be rewritten to natively support the `hm_files` section

PR #589, opened 2022-12-09
Renaming Elasticsearch indices, and the composite index name

kburke:


yuanzhou (approved):
Will merge into `main` directly since just config example update

PR #597, opened 2022-12-20
Remove mapped_specimen_type and update tests

yuanzhou:
Following merged PR #596 but for TEST
The same changes have been merged and running on DEV currently.


yuanzhou:
From John:
>it shouldn't make a difference with the data Bill created on TEST. The mapped_specimen_type code only runs if the specimen_type key is present in the ES document.

PR #607, opened 2023-01-09
John conroy/map sample category

john-conroy:
I manually updated the enums in `src/search-schema/data/definitions/enums/tissue_sample_types.yaml` to include the missing mappings for `sample_category` and added similar mapping logic previously used for `specimen_type` for `sample_category`. In this case it only capitalizes the strings.

I'm less sure of the impact on the changes in `src/hubmap_translation/indexer_base.py` and `src/hubmap_translator.py`.

### DIFF
--- a/src/hubmap_translation/addl_index_transformations/portal/__init__.py
+++ b/src/hubmap_translation/addl_index_transformations/portal/__init__.py
@@ -112,6 +112,7 @@ def transform(doc, batch_id='unspecified'):
      'ancestor_counts': {'entity_type': {}},
      'ancestor_ids': ['1234', '5678'],
      'ancestors': [{'created_by_user_displayname': 'Daniel Cotter',
+                    'mapped_sample_category': 'Section',
                     'sample_category': 'section'}],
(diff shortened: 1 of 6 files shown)


DerekFurstPitt (approved):
Regarding the changes to the enums yamls, the reason we've done things the way we did with not adding the new sample category stuff to these yamls is because, to my understanding, we're going to be doing away with  tissue_sample_types.yaml and the rest of the yamls when those eventually get handled by the ontology service. Which is why for things like display subtype and mapped sample category, we didn't point them there. 

From what I can see, there's nothing wrong with these changes and I don't see anything that will cause issues so I'm approving this, but if those yamls do end up getting removed, we'll need to obviously point these functions that rely on them somewhere else

yuanzhou (approved):
The changes look good to me, will get it out on DEV with a full reindex to verify. @john-conroy I'll let you know when things are ready on DEV for your review. Afterwards, I'll create a new PR from your branch to `main` for TEST and PROD use. No additional actions required from you. Just so you know the workflow.

PR #611, opened 2023-01-19
Completely remove specimen type

yuanzhou:
Since the `sample_category` is working fine, the old `specimen_type` and `specimen_type_other` need to be removed from ES completely.


yuanzhou:
From Bruce:
>we have removed the use of sample_category, so we shouldn't be a stopper any longer.

yuanzhou:
John Conroy confirmed that the portal-ui has completed migrated to the new filed.

PR #618, opened 2023-01-26
Revise scroll-search endpoint to use composite index names

kburke:
Revise scroll-search endpoint to use composite index names from search-config.yaml rather than OpenSearch index names.

For search-api#609, supporting files-api#9.


yuanzhou (approved):
@kburke I'll take care of the submodule reference update once the corresponding PR https://github.com/dbmi-pitt/search-adaptor/pull/21 gets merged to the `dev-integrate` branch.

PR #660, opened 2023-04-18
Welling/add publication ancillary

jswelling:
This PR adds the publication_ancillary 'assay type' to assay_types.yaml, on the road to support for Publications.


jswelling:
Drat, should have been merged against dev-integrate

PR #685, opened 2023-06-15
Add new alt names for DESI / WGS

NickAkhmetov:
Related to slack conversation: https://hubmapconsortium.slack.com/archives/C011RF3NQ23/p1686765942747269?thread_ts=1686750725.282249&cid=C011RF3NQ23

### DIFF
--- a/src/search-schema/data/definitions/enums/assay_types.yaml
+++ b/src/search-schema/data/definitions/enums/assay_types.yaml
@@ -80,14 +80,14 @@ DART-FISH:
 
 DESI:
   description: DESI
-  alt-names: [DESI-IMS]
+  alt-names: ['DESI-IMS', 'DESI IMS']
   primary: true
   contains-pii: false
   vitessce-hints: []
 
 DESI_pyramid:
   description: DESI [Image Pyramid]
-  alt-names: [["DESI", "Image Pyramid"], ["DESI", "image_pyramid"], ["DESI-IMS", "Image Pyramid"], ["DESI-IMS", "image_pyramid"]]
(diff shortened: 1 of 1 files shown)


NickAkhmetov:
> Is this PR to address some of the issues I posted from the reindex errors?
Yes, since these appeared to be close lexical matches for existing terms with either initializations or the use of dashes vs. spaces, I assumed that these could be included as synonyms; there are a few other existing cases in the assays yaml file where this is already the case. If this assumption is incorrect, I'm happy to drop this PR 😄 

> I remember you mentioned some portal transformation logic problems, have you come up with a solution?
We discovered the source of the regression with metadata tables and have fixed it on the front-end. 

> any other issues regarding the deprecated fields?
We did not find any further data-related issues during QA testing on Friday, and will perform another round of QA testing for 0.73.0.

lchoy:
> @lchoy I remember you mentioned some portal transformation logic problems, have you come up with a solution?

We still need to perform the cleanup for the portal transformation logic, but the issues we were seeing were actually related to other changes. These have been addressed in portal-ui 0.73.0.

NickAkhmetov:
From my perspective, since the yaml file is in use right now, it makes sense to keep it up to date until the indexing process is updated to use the UBKG ontology and these changes were made to dismiss the indexing errors you had observed. With that said, it's fully up to your discretion if you'd prefer to take that path forward; no feelings will be hurt if this PR is closed instead 😄

AlanSimmons:
I will update the source for the ontology neo4j instance today to add alt names for the DESI and WGS assay types.

NickAkhmetov:
I opened another PR against this branch (#689) to cover the changes to `vitessce-hints` for `mibi deepcell` assays/avoid conflicts.

yuanzhou:
@NickAkhmetov let's move forward with this PR for now. I'll get it merged and redeployed on DEV.

AlanSimmons:
By the way, the ontology neo4j that either was already deployed to production or will be very soon should reflect these changes.

yuanzhou (approved):

PR #700, opened 2023-07-26
Add group_name to configuration for all entities

kburke:
Add group_name to configuration for all entities, which will cover Uploads.  Discard block of code which would remove and re-insert entity properties during translation, since property names are normalized now.


yuanzhou:
FYI @kburke we'll test on DEV first before merging to `main`

yuanzhou (approved):

PR #701, opened 2023-07-26
Add group_name to configuration for all entities, which will cover Up…

yuanzhou:
Following merged PR #700 for TEST


yuanzhou:
Verified on DEV and all good

PR #710, opened 2023-08-18
Update Portal-Visualization to 0.0.13

yuanzhou:
Following merged PR #707 for TEST/STAGE/PROD

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -5,7 +5,7 @@ jsonschema==3.2.0
 requests==2.27.1
 PyYAML==5.4.1
 
-git+https://github.com/hubmapconsortium/portal-visualization.git@0.0.12#egg=portal-visualization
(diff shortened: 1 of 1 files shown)


yuanzhou:
The corresponding PR of portal-ui had been merged and released ahead of this one. 

According to Nikolay:
>We're in the clear as far as this update is concerned, but I will make sure that any future updates are synchronized.
The has_visualization function takes a hubmap entity and an assay lookup function, then runs it through the get_view_config_builder function, which analyzes the vitessce hints/dag names associated with the entity's assays and returns the appropriate view conf builder. The has_visualization function then checks if the returned builder is a real builder or a null conf builder, which is the return for datasets without defined visualizations.
Since the updates that were made to portal_visualization did not affect the logic for which builder to use, only the output of the builders, deploying this update to portal-ui before the search-api changes were merged did not cause any regressions in this case.

PR #728, opened 2023-12-13
Abandon YAML for indexing, using UBKG

kburke:
Remove YAML file usage for organ Description retrieval. Replace tissue type info from YAML with sample_category for non-organ Samples

### DIFF
--- a/src/hubmap_translator.py
+++ b/src/hubmap_translator.py
@@ -49,6 +49,9 @@ class Translator(TranslatorInterface):
     ACCESS_LEVEL_PUBLIC = 'public'
     ACCESS_LEVEL_CONSORTIUM = 'consortium'
     DATASET_STATUS_PUBLISHED = 'published'
+    # Base URL and constants to build endpoint URLs for talking to Ontology API
+    ONTOLOGY_API_BASE_URL = 'https://ontology-api.dev.hubmapconsortium.org'.strip('/')
+    ONTOLOGY_API_ORGAN_TYPES_ENDPOINT = '/organs/by-code?application_context=HUBMAP'
     DEFAULT_INDEX_WITHOUT_PREFIX = ''
     INDICES = {}
     TRANSFORMERS = {}
@@ -57,7 +60,6 @@ class Translator(TranslatorInterface):
     skip_comparision = False
     failed_entity_api_calls = []
     failed_entity_ids = []
-    _ontology_api_dict = None
 
     def __init__(self, indices, app_client_id, app_client_secret, token, ontology_api_dict:dict=None):
         try:
@@ -72,7 +74,6 @@ class Translator(TranslatorInterface):
             self.DEFAULT_INDEX_WITHOUT_PREFIX: str = indices['default_index']
             self.INDICES: dict = {'default_index': self.DEFAULT_INDEX_WITHOUT_PREFIX, 'indices': self.indices}
             self.DEFAULT_ENTITY_API_URL = self.INDICES['indices'][self.DEFAULT_INDEX_WITHOUT_PREFIX]['document_source_endpoint'].strip('/')
-            self._ontology_api_dict = ontology_api_dict
 
             self.indexer = Indexer(self.indices, self.DEFAULT_INDEX_WITHOUT_PREFIX)
 
@@ -793,7 +794,7 @@ class Translator(TranslatorInterface):
     # Upload: Just make it "Data Upload" for all uploads
     # Donor: "Donor"
     # Sample: if sample_category == 'organ' the display name linked to the corresponding description of organ code
-    # otherwise sample_category code as the display name for Block, Organ, or Suspension.
+    # otherwise sample_category code as the display name for Block, Section, or Suspension.
     # Dataset: the display names linked to the values in data_types as a comma separated list
     def generate_display_subtype(self, entity):
         logger.info("Start executing generate_display_subtype()")
@@ -1179,11 +1180,7 @@ class Translator(TranslatorInterface):
         }
     """
     def get_organ_types(self):
-        # Count on good values in the ontology dict from error checking in main.py
-        ontology_api_url = self._ontology_api_dict['ONTOLOGY_API_BASE_URL']
-        ontology_api_app_context = self._ontology_api_dict['ONTOLOGY_API_APP_CONTEXT']
-
-        target_url = f"{ontology_api_url}/organs/by-code?application_context={ontology_api_app_context}"
+        target_url = f"{self.ONTOLOGY_API_BASE_URL}{self.ONTOLOGY_API_ORGAN_TYPES_ENDPOINT}"
 
         # Disable ssl certificate verification, and use the read-only ontology-api without authentication.
         response = requests.get(url=target_url, verify=False)
--- a/src/hubmap_translator.py
+++ b/src/hubmap_translator.py
@@ -57,9 +57,9 @@ class Translator(TranslatorInterface):
     skip_comparision = False
     failed_entity_api_calls = []
     failed_entity_ids = []
+    _ontology_api_dict = None
 
-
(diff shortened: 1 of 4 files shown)


src/hubmap_translator.py
```diff
@@ -792,7 +793,7 @@ def add_calculated_fields(self, entity):
     # Upload: Just make it "Data Upload" for all uploads
     # Donor: "Donor"
     # Sample: if sample_category == 'organ' the display name linked to the corresponding description of organ code
-    # otherwise the display name linked to the value of the corresponding description of sample_category code
+    # otherwise sample_category code as the display name for Block, Organ, or Suspension.
```
yuanzhou, line 796: Can we add the missing "Section" to the comment? Just to be complete.

src/hubmap_translator.py
```diff
@@ -57,9 +57,9 @@ class Translator(TranslatorInterface):
     skip_comparision = False
     failed_entity_api_calls = []
     failed_entity_ids = []
+    _ontology_api_dict = None
 
-
-    def __init__(self, indices, app_client_id, app_client_secret, token):
+    def __init__(self, indices, app_client_id, app_client_secret, token, ontology_api_dict:dict=None):
```
yuanzhou, line 62: @kburke I understand you made the application context configurable, there's really no need to do that, it'll be a fixed value `application_context=HUBMAP`. This is how the UBKG ontology-api hardcoded internally for HuBMAP and SenNet.

Passing in a single URL base is sufficient and easier.

src/hubmap_translator.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, indices, app_client_id, app_client_secret, token):
             self.DEFAULT_INDEX_WITHOUT_PREFIX: str = indices['default_index']
             self.INDICES: dict = {'default_index': self.DEFAULT_INDEX_WITHOUT_PREFIX, 'indices': self.indices}
             self.DEFAULT_ENTITY_API_URL = self.INDICES['indices'][self.DEFAULT_INDEX_WITHOUT_PREFIX]['document_source_endpoint'].strip('/')
+            self._ontology_api_dict = ontology_api_dict
```
yuanzhou, line 75: When assigning the ontology-api URL, please also strip out trailing slash to avoid double `//` being composed later in the actual call, which will be treated as Not Found by the API Gateway.

src/hubmap_translator.py
```diff
@@ -1154,7 +1160,51 @@ def delete_and_recreate_indices(self):
...
+            "HT": "Heart",
+            ...
+        }
+    """
+    def get_organ_types(self):
+        # Count on good values in the ontology dict from error checking in main.py
+        ontology_api_url = self._ontology_api_dict['ONTOLOGY_API_BASE_URL']
+        ontology_api_app_context = self._ontology_api_dict['ONTOLOGY_API_APP_CONTEXT']
+
+        target_url = f"{ontology_api_url}/organs/by-code?application_context={ontology_api_app_context}"
```
yuanzhou, line 1186: This is where double slashes may happen if the config includes a trailing slash.

yuanzhou (changes requested):
@kburke In addition to the minor tweaks described below. Can you also add the new config to the `app.cfg.example`?

src/hubmap_translator.py
kburke: I was passing around the application_context, which I understand must be a part of the ontology-api URL for the first two reasons that follow:
- Initially I was trying to make this work as a replacement **_inside_** search-adaptor's [src/translator/tranlation_helper_functions.py](https://github.com/dbmi-pitt/search-adaptor/blob/main/src/translator/tranlation_helper_functions.py), which meant SN would be using the same code.  If you affirm me "pulling this up" into search-api, I can inline the `application_context` rather than read it from configuration.
- But if I inline the code, I will change [this line](https://github.com/hubmapconsortium/search-api/blob/a72bd216019916d4e71e0ceb0ea3bc5ec3ef9154/src/hubmap_translator.py#L1186) to be
`target_url = f"{ontology_api_url}/organs/by-code?application_context=HUBMAP"`, right?  And that will mean SN needs to notice it there and set it, rather than configuration.  Just want to confirm that is preferred for one-time, immutable settings...
- And is the URL any less immutable?  Should I inline that, to really simplify?

src/hubmap_translator.py
kburke: @yuanzhou My typo, sorry, but just to confirm, the list is **_only_** Block, Section, or Suspension.  Organ should **_not_** be in the list, as that is previous line documents the return value will be an organ Description from UBKG.

src/hubmap_translator.py
yuanzhou: @kburke I appreciate the thinking you putting into the work for compatibility considerations. In case I didn't fully understand you or maybe you didn't understand me, let's sort this out:
- SenNet doesn't use this `get_type_description()` from the helper functions, so it's totally fine the way you did by removing it: https://github.com/dbmi-pitt/search-adaptor/pull/50/commits/7b8e9121142789e6143ec78f42248340b75f510b
- Yes, we'll be using `target_url = f"{ontology_api_url}/organs/by-code?application_context=HUBMAP"` in the HuBMAP search-api code, and SenNet should not care about this. Consider SenNet search-api repo a totally different application.
- I treated the URL path as a constant in entity-api: https://github.com/hubmapconsortium/entity-api/blob/main/src/schema/schema_constants.py#L18 to minimize the actual code changes. In search-api, you can also define it at the top to assist the future change (unlikely): https://github.com/hubmapconsortium/search-api/blob/karlburke/removeYAMLuseUBKG/src/hubmap_translator.py#L49-L51

kburke: Thanks, @yuanzhou , I understand and will get it in that way.

src/hubmap_translator.py
kburke: Committed without Organ in list, but can change tomorrow, if appropriate.

src/hubmap_translator.py
kburke: Committed with these constants embedded where Joe suggests above, inside `hubmap_translator.py`.
This removes the need to commit `app.cfg.example`, as request at the top

src/hubmap_translator.py
kburke: Done, adjacent to constant declaration in `hubmap_translator.py`.

src/hubmap_translator.py
kburke: Reformatted line setting `target_url` to use new constants.

src/hubmap_translator.py
```diff
@@ -49,6 +49,9 @@ class Translator(TranslatorInterface):
     ACCESS_LEVEL_PUBLIC = 'public'
     ACCESS_LEVEL_CONSORTIUM = 'consortium'
     DATASET_STATUS_PUBLISHED = 'published'
+    # Base URL and constants to build endpoint URLs for talking to Ontology API
+    ONTOLOGY_API_BASE_URL = 'https://ontology-api.dev.hubmapconsortium.org'.strip('/')
```
yuanzhou, line 53: @kburke there still must be confusions. We'll only need the base URL configurable, that's really what meant. DEV/TEST uses the DEV instance of ontology-api and STAGE/PROD uses the PROD instance. We can't hardcode the base URL.

yuanzhou (changes requested):

src/hubmap_translator.py
yuanzhou: @kburke we need this base URL configurable in `app.cfg.example`

kburke: @yuanzhou Returned the base URL to the app.cfg.example, left endpoint URL constant where I understand the 3rd comment of this chain suggests it be placed.

yuanzhou (approved):

PR #729, opened 2023-12-13
John conroy/soft assay translations

john-conroy:
This PR updates our portal transformation code to no longer depend on the `data_types` field. 

For the time being we are still representing the human readable assay term, `description` from the soft assay endpoint, as `mapped_data_types` in the entity document. ~This will make the immediate transition smoother and we'll update the field with a more accurate name following the break.~ Update 1/2: I've now added `assay_display_name`. After we get these changes to PROD, we'll switch to using `assay_display_name` in the portal and can remove `mapped_data_types` here.

This is only configured to work on `DEV`, @yuanzhou once we confirm this logic works on `DEV` will you configure for each env?


lchoy (commented):
Could we move the `search-api/src/hubmap_translation/addl_index_transformations/portal/tests` folder to the same level as `search-api/src`, pulling it out of the application code?

yuanzhou:
@john-conroy thanks for the changes, I've also changed the base branch to `dev-integrate` for this PR to follow the workflow. Yes, I'll take care of the config for each environment after we finish DEV testing and move forward to TEST/STAGE/PROD.

>Could we move the search-api/src/hubmap_translation/addl_index_transformations/portal/tests folder to the same level as search-api/src, pulling it out of the application code?

Yes, we can do that. If we do so, will also need to update the test script used by Travis CI: https://github.com/hubmapconsortium/search-api/blob/john-conroy/soft-assay-translations/.travis.yml#L22

john-conroy:
Thanks @yuanzhou! I think we can defer the testing changes to get this in sooner.

src/hubmap_translation/addl_index_transformations/portal/translate.py
```diff
@@ -23,7 +20,7 @@ def _unexpected(s):
 
 
 def translate(doc):
-    _add_has_visualization(doc)
+    # _add_has_visualization(doc)
```
NickAkhmetov, line 23: Could you clarify the reason for the removal of the `add_has_visualization` functionality? The ingest API endpoint should be up in all environments.

john-conroy: I just forgot to delete the commented out line. The `visualization` bool is added in src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py.

NickAkhmetov (approved):
Looks good to me! Thanks for the clarification.

lchoy (approved):

john-conroy:
@yuanzhou could you re-index DEV with these changes at your convenience?

yuanzhou (approved):

PR #731, opened 2023-12-18
added dataset_types (publications, datasets) to data/definitions/fields.yaml

DerekFurstPitt:
added dataset_types (publications, datasets) to data/definitions/fields.yaml

### DIFF
--- a/src/search-schema/data/definitions/fields.yaml
+++ b/src/search-schema/data/definitions/fields.yaml
@@ -59,6 +59,13 @@ data_types:
   - publication
   enum: assay_types
   required: false
+dataset_type:
+  description: 'The assay type contained in the Dataset or Publication.'
+  entity_types:
+  - dataset
+  - publication
+  enum: assay_types
+  required: true
 description:
   description: A description of the Entity, hand entered into the UI
   entity_types:


yuanzhou:
@john-conroy does the portal index transformation use this `src/search-schema/data/definitions/fields.yaml`? If so, how? We've deprecated a few yaml files under `src/search-schema/data/definitions/enums/`:
- `tissue_sample_types.yaml`
- `organ_types.yaml`
- `assay_types.yaml`

I'm wondering if we should consolidate the rest of the yamls including this `fields.yaml` in the parent directory. Also FYI @shirey @AlanSimmons @lchoy if you have any insights.

john-conroy:
I don't believe we depend on `fields.yaml`, but I'm checking now.

We still depend on `data_access_levels`, `dataset_status_types`, `organ_types`, and `tissue_sample_types`. Of note for `tissue_sample_types`, we are solely using the mappings for the values in the `sample_category` field which are just capitalizations. 

We can create a plan to switch these over to UBKG(?) after the winter break.

AlanSimmons:
1. `organ_types` has been modeled in the UBKG, and endpoints for organ infromation available in the UBKG API via the [organs](https://smart-api.info/ui/d10ff85265d8b749fbe3ad7b51d0bf0a#/default/get_organ_types) routes.
2. `tissue_sample_types`, as I understood from Stephen Fisher, was to be replaced completely by information from the CEDAR integration. This file is not modeled in the UBKG.
3. The UBKG does not currently model `data_access_levels`,`dataset_status_types`, or `entity_types` for HuBMAP; however, it does model the equivalents for SenNet. It would be straightforward to model the HuBMAP files in UBKG. SenNet did not ask for UBKG API endpoints, but they could be developed.

john-conroy:
> tissue_sample_types, as I understood from Stephen Fisher, was to be replaced completely by information from the CEDAR integration. This file is not modeled in the UBKG.

We can likely just remove the map and capitalize the string ourselves.

yuanzhou:
> > tissue_sample_types, as I understood from Stephen Fisher, was to be replaced completely by information from the CEDAR integration. This file is not modeled in the UBKG.
> 
> We can likely just remove the map and capitalize the string ourselves.

I would do that instead of calling UBKG. The `sample_category` returned from entity-api is basically just one of the 4 values: organ, block, section, and suspension.

yuanzhou:
@AlanSimmons and @john-conroy WRT `data_access_levels`, `dataset_status_types`, and `entity_types`, I'd consider them as constants and they are stable in HuBMAP. No need to model them in UBKG to introduce extra steps when we can just define them as constants in search-api for the portal needs.

yuanzhou:
Closing this PR to favor the deprecation. The actual field `dataset_type` had been released and it's showing up in the indices.

PR #767, opened 2024-03-18
Fixes "IndexError: list index out of range" in add_partonomy.py.

lchoy:
Fixes "IndexError: list index out of range" in add_partonomy.py when the `origin_samples` field is not available.
Fixes CAT-529.

PR #770, opened 2024-03-21
Lchoy/fix missing sample

yuanzhou:
Following merged PR #767 #771  #775 for TEST


john-conroy:
@lchoy what motivated the change between this and the original PR?

lchoy:
> @lchoy what motivated the change between this and the original PR?

Since the Portal UI is unable to handle the missing organ details, this PR changes the behavior so that the document is not indexed the the Portal index. Previously, the API was throwing an IndexError. Now, a message is logged to explain that this entity is invalid due to the missing data.

john-conroy (approved):

PR #775, opened 2024-04-04
Catch IndexError when a sample or dataset is missing an organ sample.

lchoy:
This PR addresses the occurrence of "IndexError: list index out of range" when a sample or dataset entity is missing associations to an organ. When this condition occurs, a "TranslationException: Invalid document" message will be logged to state that the document is "Missing or empty 'origin_samples'." The entity will not be added to the portal index to prevent issues on the UI. These entities will be viewable in the ingest portal or data dashboard.

### DIFF
--- a/src/hubmap_translation/addl_index_transformations/portal/tests/test_add_partonomy.py
+++ b/src/hubmap_translation/addl_index_transformations/portal/tests/test_add_partonomy.py
@@ -11,7 +11,7 @@ from hubmap_translation.addl_index_transformations.portal.translate import Trans
             {}, None, id="empty doc does not throw exception"
         ),
         pytest.param(
-            {"uuid": "test_dataset_uuid", "entity_type": "Donor"}, None, id="Donor is expected to have missing origin_samples"
+            {"uuid": "test_uuid", "entity_type": "Donor"}, None, id="Donor is expected to have missing origin_samples"
         ),
         pytest.param(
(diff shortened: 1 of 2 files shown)


src/hubmap_translation/addl_index_transformations/portal/tests/test_add_partonomy.py
```diff
@@ -1,20 +1,31 @@
...
 
 @pytest.mark.parametrize(
     "doc, expected_organ_iri",
     [
         pytest.param(
-            {"origin_samples": [{"organ": "UT"}]}, None, id="valid organ"
+            {}, None, id="empty doc does not throw exception"
+        ),
+        pytest.param(
+            {"uuid": "test_dataset_uuid", "entity_type": "Donor"}, None, id="Donor is expected to have missing origin_samples"
```
NickAkhmetov, line 14:
```suggestion
            {"uuid": "test_uuid", "entity_type": "Donor"}, None, id="Donor is expected to have missing origin_samples"
```

Small idea to make this a bit clearer at a glance 👍🏻

NickAkhmetov (approved):
Just one idea to potentially make the test case clearer; looks great otherwise!

yuanzhou (approved):
@lchoy I'll do a full reindex on DEV overnight and we'll find out tomorrow.

Issue #756, opened 2024-03-05
Switch to the specialized entity-api call `GET /documents/<id>`


yuanzhou:
**AFTER** implementing the new features described in entity-api via https://github.com/hubmapconsortium/entity-api/issues/630, we'll need to replace the current entity-api call with this new `GET /documents/<id>` in `hubmap_translator.py` when retrieving the individual entity details. No changes to other entity-api calls (ancestors/descendants/revisions...).

PR #777, opened 2024-04-09
Initial revisions for search-api/issues/756. Use of new entity-api /documents endpoint

kburke:
Initial revisions for search-api/issues/756. Use of new entity-api /documents endpoint to create indexable documents rather than the /entities endpoint. Approve indexing efficiency by not generating fields which will be removed before inserting documents into OpenSearch indices.


yuanzhou:
Link to issue https://github.com/hubmapconsortium/search-api/issues/756

yuanzhou (approved):

PR #779, opened 2024-04-15
Initial, companion revisions for entity-api/issues/631

kburke:
Initial revisions for entity-api/issues/631. Move calculation of display_subtype from search-api to entity-api to take advantage of caching for indexing efficiency. Companion revision in entity-api adds display_subtype.


src/hubmap_translation/neo4j-to-es-attributes.json
```diff
@@ -206,6 +206,9 @@
     },
     "superseded_associated_processed_component_uuids": {
       "es_name": "superseded_associated_processed_component_uuids"
+    },
+    "display_subtype": {
```
yuanzhou, line 210: @kburke thanks for taking this into consideration for this round of updates. Soon we'll refactor out the usage of this mapping json file and use the entity-api schema yaml flag `indexed: false` to exclude the ones that won't be indexed.

yuanzhou (approved):

PR #794, opened 2024-05-06
Bump portal-vis to 0.2.2

NickAkhmetov:
This PR is a re-creation of #792, based on main instead of dev-integrate.


yuanzhou:
Will merge this PR after PR #796 gets tested on DEV.

yuanzhou:
Will hear back from @NickAkhmetov 
>once we confirm full functionality on DEV (i.e. once there's a newly processed dataset with which we can confirm portal-vis is working as expected)

yuanzhou (approved):

PR #804, opened 2024-05-21
NickAkhmetov/CAT-665 Adjust visualization flag logic to check descendants

NickAkhmetov:
This PR:
- Bumps portal-visualization to 0.2.5 (to pull in changes to the `has_visualization` function to allow usage of parent UUIDs)
- Adjusts the `visualization` flag logic so that it performs vislifting to take into account whether the dataset has supporting image pyramid entities which can be visualized.

This change will require the addition of the entity API descendants URL to the search_config.yaml for each environment (`ingest_api_descendants_url`, following the convention set by the soft assay URL).

### DIFF
--- a/src/instance/search-config.yaml.example
+++ b/src/instance/search-config.yaml.example
@@ -4,6 +4,9 @@
 # Soft assay support via ingest-api GET /assaytype/<dataset-id> endpoint
 # Omit the <dataset-id> in this config 
 ingest_api_soft_assay_url: https://ingest-api.dev.hubmapconsortium.org/assaytype/ 
+# Descendants support via ingest-api GET /descendants/<dataset-id> endpoint
+# Omit the <dataset-id> in this config 
+ingest_api_descendants_url: https://ingest-api.dev.hubmapconsortium.org/descendants/ 
 
 # default index name for endpoints that don't specify an index
 default_index: entities
--- a/src/instance/search-config.yaml.example
+++ b/src/instance/search-config.yaml.example
@@ -4,10 +4,6 @@
 # Soft assay support via ingest-api GET /assaytype/<dataset-id> endpoint
 # Omit the <dataset-id> in this config 
 ingest_api_soft_assay_url: https://ingest-api.dev.hubmapconsortium.org/assaytype/ 
-# Descendants support via ingest-api GET /descendants/<dataset-id> endpoint
-# Omit the <dataset-id> in this config 
-ingest_api_descendants_url: https://ingest-api.dev.hubmapconsortium.org/descendants/ 
-
 # default index name for endpoints that don't specify an index
 default_index: entities
 
--- a/src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py
+++ b/src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py
@@ -140,9 +140,8 @@ def add_assay_details(doc, transformation_resources):
 
         # Check if the main entity can be visualized by portal-visualization.
         has_viz = has_visualization(doc, get_assay_type_for_viz)
-        if has_viz:
-            doc['visualization'] = True
-        else:
+        doc['visualization'] = has_viz
+        if not has_viz:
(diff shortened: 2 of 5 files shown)


src/instance/search-config.yaml.example
```diff
@@ -4,6 +4,9 @@
 # Soft assay support via ingest-api GET /assaytype/<dataset-id> endpoint
 # Omit the <dataset-id> in this config 
 ingest_api_soft_assay_url: https://ingest-api.dev.hubmapconsortium.org/assaytype/ 
+# Descendants support via ingest-api GET /descendants/<dataset-id> endpoint
+# Omit the <dataset-id> in this config 
+ingest_api_descendants_url: https://ingest-api.dev.hubmapconsortium.org/descendants/
```
yuanzhou, line 9: @NickAkhmetov There's no descendants endpoint in ingest-api. It's available via entity-api `GET /descendants/<id>`. And if you foresee there could be a large number of descendants being returned, you can use query string `GET /descendants/<id>?property=uuid` to return only the uuids rather than the complete details which will likely cause timeout. Afterwards, call entity-api `/documents/<id>` to retrieve the target dataset detail.

The entity-api base URL is already there: https://github.com/hubmapconsortium/search-api/blob/d50ddaae79dc3a1b6d95a37e6c0acaabfe1c5c15/src/instance/search-config.yaml.example#L32

yuanzhou (changes requested):
@NickAkhmetov please also fix the broken CI test.

src/instance/search-config.yaml.example
NickAkhmetov: > And if you foresee there could be a large number of descendants being returned, you can use query string `GET /descendants/<id>?property=uuid` to return only the uuids rather than the complete details which will likely cause timeout. Afterwards, call entity-api `/documents/<id>` to retrieve the target dataset detail.

@yuanzhou Given that the vislifting logic needs to find at least one support entity with a `published` or `QA` status, I am worried that separately requesting the data for each of those documents would pose a different performance concern and severely slow down indexing.

---

For clarity: `n` is the total number of descendants of the parent entity, `m` is the number of published/QA descendants of the parent entity, and `p` is the number of image pyramid support entities.

---

The current approach is optimized around minimizing the number of network calls for each dataset; in a best case scenario, a total of only 3 network calls are made, which stays relatively constant:

1. Check if the parent entity can be visualized (1 network call to the soft assay endpoint)
2. request the full descendants list (1 network call to the descendants endpoint)
3. filter out non-published/non-QA descendants and sort them in order of recency 
   a. no network calls, local operations with complexity `O(n)` and `O(mlogm)` respectively.
4. request the assaytypes for the remaining descendants, short circuiting with a true return if any of them are a supporting image pyramid

In total, m + 2 network calls are made.

---

If we adjust this logic to look up only the uuids during step 2, we lose the optimizations that filtering by status and sorting by last modified date were giving us:

1. Check if the parent entity can be visualized (still 1 network call)
2. Request the full list of descendant UUIDs (1 network call)
3. For each descendant, request the soft assay information (n network calls)
4. Filter out non-image pyramids based on the results from the soft assay lookups
5. Look up the full information for the remaining descendants and select the most recent published one (p network calls, where p is the number of image pyramid support datasets)

In total, n + p + 2 network calls would need to be made in this case, which is guaranteed to be greater than or equal to the number of calls made by the current approach since `n` is a superset of `m`.

---

If we are worried about getting too large of a response from the descendants endpoint, what would the lift be to enhance the descendants endpoint to accept multiple property keys, e.g. something like `/descendants/<uuid>?property=uuid?property=last_modified_timestamp?property=status`, or to spin up a corresponding `POST` endpoint which would accept the requested properties from the body of the request?

yuanzhou (approved):

NickAkhmetov:
@yuanzhou Thanks for the approval! We'll end up iterating on this solution if an appropriate vitessce hint is added for the image pyramid parents, which would allow us to skip the vislifting process entirely during indexing.

PR #806, opened 2024-05-23
Revert "NickAkhmetov/CAT-665 Adjust visualization flag logic to check descendants"

NickAkhmetov:
This change should have gone into `dev-integrate` rather than directly into main


yuanzhou:
Will sync main branch into dev-integrate and test on DEV first. Won't use this revert PR this time.

PR #813, opened 2024-06-10
Karlburke/indexing efficiencies round5

kburke:
New indexing efficiencies - update OpenSearch documents in the `entities` indices rather than re-index them.


yuanzhou (approved):
Will test thoroughly on DEV using a new set of indices

PR #818, opened 2024-06-20
Karlburke/indexing efficiencies round5

kburke:
Refactoring of call_indexer() into specific methods each encapsulating a different functionality previously determined by the combination of arguments provided.


yuanzhou (approved):
Will do more testing on DEV

PR #820, opened 2024-06-24
Patch new reindexing behavior to be threaded, as previously implemented.

kburke:
Fix per PR title to remediate severe performance impact from single-threaded implementation.


yuanzhou (approved):
Will test on DEV and find out the total reindex time change

PR #826, opened 2024-07-01
Karlburke/indexing efficiencies round6

yuanzhou:
Following merged PR #824 for TEST

Will need to go hand in hand with https://github.com/hubmapconsortium/entity-api/pull/696


yuanzhou:
Replaced by PR #829 

FYI @kburke

PR #827, opened 2024-07-02
Corrected incorrect indexing caused by misidentifying when relationships change

kburke:
Corrected incorrect indexing caused by misidentifying when relationships in Neo4j had changed by delay reindex of entity until after the relationship evaluation.


kburke:
This branch contains commits already in dev-integrate.  Closing without merge.

PR #837, opened 2024-07-18
Use ElasticSearch's retry_on_conflict feature on _update

kburke:
Initial commit using ElasticSearch's `retry_on_conflict` feature on _update operations.


kburke:
@yuanzhou Pushed a commit to move configurable `retry_on_conflict` setting from `app.cfg.example` to `search-config.yaml.example`.

yuanzhou (approved):
Thanks @kburke ! Will get this merged into `main` directly and find out on PROD. Consider this as a hot fix so we don't follow the regular DEV->TEST->RPOD flow.

PR #844, opened 2024-07-24
Cascade Dataset updates to Collection and Upload

kburke:
Cascade Dataset updates to also update Collection and Upload documents containing the Dataset.


src/hubmap_translator.py
```diff
@@ -567,12 +562,32 @@ def translate(self, entity_id):
                                                             , endpoint_suffix=None
                                                             , url_property='uuid')
 
+                # If working with a Dataset, it may be copied into ElasticSearch documents for
+                # Collections and Uploads, so identify any of those which must be reindexed.
+                neo4j_collection_ids = []
+                neo4j_upload_ids = []
+                if entity['entity_type'] == 'Dataset':
```
yuanzhou, line 569: @kburke We should also include `Publication` for this use case because `Publication` is a subclass of `Dataset`. Currently we are handling the `Publication-Collection` relationship differently from `Dataset-Collection` in Neo4j, and this may change down the road.

yuanzhou (changes requested):

src/hubmap_translator.py
kburke: @yuanzhou Requested change pushed to branch, thanks!

yuanzhou (approved):

PR #854, opened 2024-08-15
Update docker-compose.yml

yuanzhou:


yuanzhou:
@kburke I'll merge this docker deployment volume mount since it's not affecting anything else.

PR #896, opened 2024-10-14
exclude properties from public indices

kburke:
Initial commit to keep fields out of ElasticSearch documents in the public indices when Entity API's provenance_schema.yaml lists the fields in the excluded_properties_from_public_response entry for an entity.


yuanzhou (changes requested):
@kburke I'll have to finish the in-depth review during your absence and get things tested on DEV. For now, can you add this `entity_api_prov_schema_raw_url` to the example yaml config file?

kburke:
Missing entry committed and pushed, @yuanzhou . Thanks!

yuanzhou (approved):

PR #906, opened 2024-11-04
John conroy/ addl soft assay fields

john-conroy:
- Adds `soft_assaytype` to the document so we no longer have to call the ingest-api from portal-ui or portal-visualization.
- Tries to get the `pipeline` from the `pipeline-shorthand` instead of depending on regex.

### DIFF
--- a/src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py
+++ b/src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py
@@ -137,8 +137,13 @@ def add_assay_details(doc, transformation_resources):
         doc['raw_dataset_type'] = re.sub(
             "\\[(.*?)\\]", '', doc.get('dataset_type', '')).rstrip()
 
-        if pipeline := re.search("(?<=\\[)[^][]*(?=])", doc.get('dataset_type', '')):
+        if pipeline := assay_details.get('pipeline-shorthand'):
+            doc['pipeline'] = pipeline
+        elif pipeline := re.search("(?<=\\[)[^][]*(?=])", doc.get('dataset_type', '')):
             doc['pipeline'] = pipeline.group()
+
+        if soft_assaytype := assay_details.get('assaytype'):
+            doc['soft_assaytype'] = soft_assaytype
         # Preserve the previous shape of mapped_data_types.
         doc['assay_display_name'] = [assay_details.get('description')]
         # Remove once the portal-ui has transitioned to use assay_display_name.
--- a/src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py
+++ b/src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py
@@ -130,6 +130,20 @@ def _get_descendants(doc, transformation_resources):
         raise
 
 
+def _add_pipeline(doc, assay_details):
+    if pipeline := assay_details.get('pipeline-shorthand'):
+        doc['pipeline'] = pipeline
+    #  pipeline-shorthand is not returned for EPICs.
+    elif doc.get('processing_type') == 'external':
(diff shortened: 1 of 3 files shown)


src/hubmap_translation/addl_index_transformations/portal/add_assay_details.py
```diff
@@ -137,8 +137,13 @@ def add_assay_details(doc, transformation_resources):
         doc['raw_dataset_type'] = re.sub(
             "\\[(.*?)\\]", '', doc.get('dataset_type', '')).rstrip()
 
-        if pipeline := re.search("(?<=\\[)[^][]*(?=])", doc.get('dataset_type', '')):
+        if pipeline := assay_details.get('pipeline-shorthand'):
+            doc['pipeline'] = pipeline
+        elif pipeline := re.search("(?<=\\[)[^][]*(?=])", doc.get('dataset_type', '')):
             doc['pipeline'] = pipeline.group()
+
```
NickAkhmetov, line 144: Could we also handle the `Image Pyramid` and `EPIC` cases in the search index? We currently have to apply some workarounds for those cases since they don't have pipeline-shorthand info: https://github.com/hubmapconsortium/portal-ui/blob/37143cf5b4af35c27f528f6ec5ae6378f30686b4/context/app/static/js/components/detailPage/DatasetRelationships/hooks.ts#L40-L58

john-conroy: Does it make sense to handle the EPIC case here?

john-conroy: The regex fallback is still in place here and image pyramids do have a `pipeline` of 'Image Pyramid' in their docs currently because of that. I can have the logic prefer 'Image Pyramid Generation' over 'Image Pyramid' here? This would update the labels in the table of contents I believe. 

https://portal.hubmapconsortium.org/browse/dataset/7e3f3b5d1890f9612efb91d3a820cdcc#section-hbm265.ljvk.224.

NickAkhmetov: Right; it'd definitely be ideal to keep the `pipeline` values intact. 
To clarify my ask, we currently rely on the `pipeline-shorthand` for the step details in the relationship diagram for non-image pyramids, which is blank for Image Pyramid and EPIC cases, so I'm wondering if it would make sense to populate that field in search to simplify the front-end logic, since it's otherwise blank for those dataset types.

NickAkhmetov: Per a slack conversation:
* `pipeline` should be the single source of truth, no need to use `pipeline-shorthand` on the frontend anymore since it's populated from that field; image pyramids will retain `image pyramid` as the description.
* EPIC `pipeline` values will be populated from the description.

john-conroy:
@NickAkhmetov to confirm adding `soft_assaytype` we should be able to update https://github.com/hubmapconsortium/portal-visualization/blob/dbcc9aeca5e5e49b2cd44d9cdb4fabd7badd60a1/src/portal_visualization/client.py#L277 and https://github.com/hubmapconsortium/portal-visualization/blob/dbcc9aeca5e5e49b2cd44d9cdb4fabd7badd60a1/src/portal_visualization/builder_factory.py#L85 to use the doc?

NickAkhmetov:
@john-conroy As far as I can tell, those usages should all be replaceable. 👍🏻

yuanzhou (approved):

yuanzhou:
@lchoy, still waiting for you approval before I merge and deploy on DEV

lchoy (approved):

PR #920, opened 2025-03-17
added missing openapi entries

DerekFurstPitt:
added open api spec yaml entries for endpoints that didn't have them. made changes throughout and added examples where appropriate

### DIFF
--- a/search-api-spec.yaml
+++ b/search-api-spec.yaml
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.0.3
 info:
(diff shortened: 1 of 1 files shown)


yuanzhou (changes requested):
@DerekFurstPitt We'll need a few tweaks:
- Use `openapi: 3.0.3`, the same we used for entity-api
- Remove `/status` and `/reindex-all`, they are meant for internal use
- Empty response example of `{}` is being used for `/search`, `/{index_name}/search
`, `/param-search/{entity_type}`, `/mget`, `/{index_without_prefix}/mget`, `/mapping`, and `/{index_without_prefix}/mapping`. 
- Rename `{index_without_prefix}` to `{index_name}` everywhere for consistency and also not to expose too much details. Basically only the canonical index names returned from `/indices`.
- Remove the mention of `hm_consortium_entities` from `/{index_without_prefix}/mget`

PR #921, opened 2025-03-18
Adds explicit mappings for timestamp fields

lchoy:
Explicitly map the `created_timestamp`, `last_modified_timestamp`, and `published_timestamp` to `long` types to prevent rounding errors when sorting. These fields were inconsistently mapped to `float` or `long` across indices.


src/hubmap_translation/addl_index_transformations/portal/config.yaml
```diff
@@ -78,6 +78,24 @@ mappings:
         mapping:
           type: keyword
 
+    # Added 3/18/2025
+    # Epoch timestamps should be treated as long fields as 
+    # defaulting to floats produces rounding errors on sort
+    - published_timestamp:
+        path_match: "*.published_timestamp"
```
john-conroy, line 85: Probably misunderstanding, but should we remove `.` to also map the non-nested timestamps?

lchoy: This syntax maps them all, including non-nested as shown in the snippet below.


image

I tested locally by removing and recreating all indices and confirming that all instances of these fields at all levels were mapped correctly to long types.

yuanzhou:
@lchoy Can you do me a favor by merging the latest `main` branch into yours? We have some recent updates made into `main` today. 

I'll plan to full reindex on DEV tonight along with other rule chain updates, so if you and @john-conroy can confirm the changes in this PR before EOB, we'll be able to verify tomorrow.

john-conroy (approved):

PR #931, opened 2025-05-21
Austenem/Add publication status

austenem:
Adds a true `has_publication` boolean to entities with any publications in their descendant tree. Required for implementation of a new "Show only datasets linked to publication" boolean facet on the search page of the portal UI.


john-conroy:
Apologies for not picking up on this sooner, but we might not need this. Can you try an approach like what's in https://github.com/hubmapconsortium/portal-ui/pull/3608/files#diff-c205d3f28e240649387cc7c6cd4db03627c50391076919dd305519aa10fe528a first and we can revisit here if that does not work?

john-conroy:
@yuanzhou We can close this PR

yuanzhou:
@austenem Up to you to decide if you want to delete your branch.

PR #965, opened 2025-12-16
NickAkhmetov/Add `is_integrated` to portal index transformations

NickAkhmetov:
This PR adds the `is_integrated` boolean to the transformation process. All externally processed data and datasets with more than 1 dataset in its ancestor tree are treated as integrated, as:

* If there's more than one ancestor dataset, the redirect is jarring/unclear.
* Externally processed datasets may have contributor sets that are different than their parents, even if they have only one parent.

Since this relies on the ancestor dataset counts, I've added this as a separate transformation called in `transform.py` after the `add_counts` transformation and adjusted the `test_transform` tests to include integrated dataset cases. I can also add doctests to the function itself if desired.

### DIFF
--- a/src/hubmap_translation/addl_index_transformations/portal/tests/test_transform.py
+++ b/src/hubmap_translation/addl_index_transformations/portal/tests/test_transform.py
@@ -172,7 +172,9 @@ input_doc_integrated_epic.update(
 expected_output_doc_integrated_epic = expected_output_doc.copy()
 expected_output_doc_integrated_epic.update(
     {
+        "creation_action": CreationAction.EPIC,
         "is_integrated": True,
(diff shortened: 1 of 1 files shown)


NickAkhmetov:
@yomatters Since we can determine that based on the `creation_action`, which we already request for the dataset page, the current approach should suffice - it'd be a value derived from a single other field so we don't gain much from indexing it and no additional information is made available by including it.

yuanzhou:
@yomatters As part of the review process, can you approve this PR if all looks good to you? 

I'll take care of the merge and deployment. Our established workflow is to get things tested on DEV and TEST first with full reindex (when necessary like this change). Then the portal team reviews and gives me the green light for PROD release.

PR #967, opened 2026-01-06
implemented job queue into search-api

DerekFurstPitt:


src/hubmap_translator.py
```diff
@@ -685,6 +691,144 @@ def _transform_and_write_entity_to_index_group(self, entity:dict, index_group:st
                     f" entity['uuid']={entity['uuid']},"
                     f" entity['entity_type']={entity['entity_type']}")
 
+    def enqueue_reindex(self, entity_id, search_queue, priority):
```
yuanzhou: Related to the renaming from https://github.com/dbmi-pitt/search-adaptor/pull/103#pullrequestreview-3631418220, use `reindex_queue`

PR #1027, opened 2026-05-13
added new status 'approval' to yaml, and any check for qa since they are treated the same

DerekFurstPitt:
added approval status to relevant schemas, as well as certain conditionals wherever QA was checked

### DIFF
--- a/search-api-spec.yaml
+++ b/search-api-spec.yaml
@@ -544,7 +544,8 @@ components:
                 - Error
                 - Hold
                 - Invalid
-              description: "One of: New|Processing|QA|Published|Error|Hold|Invalid"
+                - Approval
+              description: "One of: New|Processing|QA|Published|Error|Hold|Invalid|Approval"
         title:
           type: string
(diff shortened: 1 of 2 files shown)


yuanzhou:
@NickAkhmetov @lchoy As mentioned earlier, we are adding this new `Dataset.status` value of "Approval" across the board, can you take a look and see if Derek's update to the portal translation code look good?

NickAkhmetov (approved):
LGTM 👍🏻

yuanzhou (approved):

lchoy:
@yuanzhou For context, how does "Approval" differ from "QA"?

yuanzhou:
@lchoy Below is a combined answer based on Bill's response, as I had the same question. 

We support two workflows:
- QA -> Published
- QA -> Approval -> Published

The original flow will still be supported because we have a lot of stuff in QA that is already out for approval and/or already approved. This new flow really doesn't differ much.

PR #1033, opened 2026-06-01
Add multi-donor demographics transformations

NickAkhmetov:
This PR:

* Adds `donor_demographics` for datasets, samples, and publications (all entities with previously associated `donor.mapped_metadata`
* This calculation relies on `donors` and falls back to `donor` in case `donors` is not available
* For all fields, the `{fieldName}_value` key includes all unique values
* For numeric fields, a min/max/mean is calculated


yuanzhou (approved):
@NickAkhmetov Thanks for implementing support for multiple donors in portal translations and for cleaning up the fallback logic.

